PR #11, opened 2022-08-08
pin key dep versions

jharris1679:


jharris1679:
Closing because this is already covered in the dev branch

PR #38, opened 2023-06-14
Fix parallelization error 'Cadet' object has no attribute '_is_file_class'

ronald-jaepel:
In a single-threaded case (and also apparently in some multi-threaded cases) the [line 160](https://github.com/fau-advanced-separations/CADET-Process/blob/29eb5cc463892b4e969dcd07f3cd9fb968bcce23/CADETProcess/simulator/cadetAdapter.py#LL160C1-L160C47)

``` Python 
if install_path.exists():
    self._install_path = install_path
    CadetAPI.cadet_path = install_path
    self.logger.info(f"CADET was found here: {install_path}")
```

sets the cadet_path for the class object CadetAPI, which in turn sets _is_file_class and _is_file. In some multi-threaded cases on windows, this change to CadetAPI isn't inherited into the spawned threads.

This PR adds a check if this assignment occurred in the local thread and, if not, sets the cadet_path.

### DIFF
--- a/CADETProcess/simulator/cadetAdapter.py
+++ b/CADETProcess/simulator/cadetAdapter.py
@@ -254,7 +254,7 @@ class Cadet(SimulatorBase):
             raise TypeError('Expected Process')
 
         if cadet is None:
-            cadet = CadetAPI()
+            cadet = self.get_new_cadet_instance()
 
         cadet.root = self.get_process_config(process)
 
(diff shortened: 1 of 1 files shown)


CADETProcess/simulator/cadetAdapter.py
```diff
@@ -258,6 +258,9 @@ def run(self, process, cadet=None, file_path=None):
 
         cadet.root = self.get_process_config(process)
 
+        if not(hasattr(cadet, "cadet_path")):
```
schmoelder, line 261: I think you can remove the outer braces.

```
if not hasattr(cadet, "cadet_path"):
```

schmoelder (changes requested):

ronald-jaepel:
Alright, will do.

CADETProcess/simulator/cadetAdapter.py
schmoelder: Oh, and maybe add a comment why this step is necessary. It seems to me this is a workaround to fix something that should actually not happen, right?

ronald-jaepel: Removed braces, added comment. 

I'm not sure if it's surprising that line 160 doesn't always work or that it does work sometimes. Therefore the comment just explains why it's there, without assigning it as a workaround. ¯\_(ツ)_/¯

ronald-jaepel:
Have rebased to dev, incorporated the changes and changed the base branch for the PR to dev.

PR #39, opened 2023-06-15
Update .gitignore to exclude the PycharmIDE .idea folder 

ronald-jaepel:
so my IDE doesn't always reset when switching to Master


ronald-jaepel:
Will reopen soon after rebase into dev.

PR #45, opened 2023-06-29
42 fit gp model on existing population from optimization problem

### LINKED ISSUE #33: Implement partial dependence plots
Partial dependence plots show the dependence between the target function (e.g. objective) and a set of ‘target’ features, marginalizing over the values of all other features (the complement features). 

For an example, see [here](https://scikit-optimize.github.io/stable/auto_examples/plots/partial-dependence-plot.html).

For this purpose, it is required to build a surrogate model (e.g. a Gaussian Process).

### LINKED ISSUE #42: Fit GP model on existing population from optimization problem
the idea is that it's better to sample the whole / relevant parameter space instead of only finding the optimal solution

- take results from OptimizationResults 
  - `.population_all` all ever evaluated parameter samples
- build GP model from sklearn

flo-schu:
Implements a Surrogate class for fitting a GP on existing data from optimizations

### implements 
- [x] fitting the GP on an existing population of samples
- [x] estimating the objectives functions
- [x] estimating the feasible objectives space
- [x] conditioning the `optimization_problem` on fixed variables
- ~~find `x0` at the edges of the problem~~

### plots
- [x] conditional
- [x] feasible space
- [x] minimum boundary
- [x] partial dependence 
- ~~overlay contour plot with true population~~

### tests
- [x] test if surrogate model breaks or not
- [x] test convergence of surrogate model to known solutions

### implementation of surrogates for different optimiztation tasks
| :zap: priority |
|----------------|
- [x] linear inequality constraints
- [ ] linear equality constraints
- [x] (pseudo) multi-objective, 
- [x] non-linear constraints

### performance
- [x] implement algorithm that finds a very good starting point for the task
- ~~find a faster method to test if the conditioned problem contains feasible space (creating a hopsy problem needs a lot of time 1/3 of total runtime)~~ This depends currently on hopsy

### code quality
- [x] Docstrings (GPT add docstrings, include warnings and errors, use numpy style, imperative style)
- [ ] refactor `condition_optimization_problem` method to a separate class that is an `OptimizationProblem`

## open questions

- actually sampled points overlaid over mean + uncertainty of GP
  GP.predict(std=True) -> mu, sigma
- contours? Lösen über fill between mit verschiedenen quantilen

## partial dependence

the function `partial_dependence` of `sklearn` returns a value that is not the actual response of the objective function but another value. This behavior is not ideal if it should be overlaid over the true objective space and the true evaluated points.

## minimum boundary plot

This plot draws the minimum boundary of the objective function over a given parameter.
There are two main approaches to achieve this

1. drawn a large number of samples and iterate over a moving window that always returns the minimum point. This is a pragmatic approach but it will be inexact
2. create a linear space over the target parameter x that should be drawn. Then optimize wrt all other parameters and find minimum. Ideally a good starting point would be that point which is close to the ones that are already close to the minimum by weighting these can be points from the surrogate or the true model evaluation (but surrogate is probably smarter, because there will be more points available)

this problem is solved by conditioning the optimization problem on one variable. This is done by fixing the conditioning variable at a specific value and substituting that value in the matrix and moving it to the RHS (`b`). 

## closes additional issues

- closes #33

### DIFF
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -150,6 +150,10 @@ class OptimizationProblem(Structure):
 
         return wrapper
 
+    def get_variable_index(self, variable):
+        var_names = self.variable_names
+        return [i for i, v in enumerate(var_names) if variable == v][0]
+
     def ensures2d(func):
         """Make sure population is ndarray with ndmin=2."""
         @wraps(func)
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -2089,6 +2089,12 @@ class OptimizationProblem(Structure):
         add_linear_equality_constraint
         linear_equality_constraint
 
+        BUG: consider implementing _linear_constraints as a dictionary.
+             when deleting and redefining constraints sequentially, this breaks
+             because the indices of constraints change. When using it as
+             a dictionary under the hood, this won't be possible and then
+             the self.linear_constraints property can return a list as always.
+             by accessing `list(self._linear_constraints.values())`
         """
         del self._linear_constraints[index]
 
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -2088,13 +2088,6 @@ class OptimizationProblem(Structure):
         --------
         add_linear_equality_constraint
         linear_equality_constraint
-
-        BUG: consider implementing _linear_constraints as a dictionary.
-             when deleting and redefining constraints sequentially, this breaks
-             because the indices of constraints change. When using it as
-             a dictionary under the hood, this won't be possible and then
-             the self.linear_constraints property can return a list as always.
-             by accessing `list(self._linear_constraints.values())`
         """
         del self._linear_constraints[index]
 
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -3478,6 +3478,13 @@ class Objective(Structure):
                     f'{self.name}_{i}'
                     for i in range(self.n_metrics)
                 ]
+
+        if len(self.evaluation_objects) > 1:
+            labels = [
+                f"{eval_obj}_{l}"
+                for l in labels
+                for eval_obj in self.evaluation_objects
+            ]
         return labels
 
     @labels.setter
@@ -3569,6 +3576,14 @@ class NonlinearConstraint(Structure):
                     f'{self.name}_{i}'
                     for i in range(self.n_metrics)
                 ]
+
+        if len(self.evaluation_objects) > 1:
+            labels = [
+                f"{eval_obj}_{l}"
+                for l in labels
+                for eval_obj in self.evaluation_objects
+            ]
+
         return labels
 
     @labels.setter
@@ -3761,6 +3776,14 @@ class MetaScore(Structure):
                     f'{self.name}_{i}'
                     for i in range(self.n_metrics)
                 ]
+
+        if len(self.evaluation_objects) > 1:
+            labels = [
+                f"{eval_obj}_{l}"
+                for l in labels
+                for eval_obj in self.evaluation_objects
+            ]
+
         return labels
 
     @labels.setter
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -577,7 +577,7 @@ class OptimizationProblem(Structure):
         """
         if len(x) != self.n_variables:
             raise CADETProcessError(
-                f'Expected {self.n_variables} value(s)'
+                f'Expected {self.n_variables} value(s), got {len(x)}'
             )
 
         x_independent = []
@@ -2555,21 +2555,21 @@ class OptimizationProblem(Structure):
 
         return flag
 
-    def transform(self, x):
-        """Transform the optimization variables from untransformed parameter space.
+    def transform(self, x_independent):
+        """Transform the independent optimization variables from untransformed parameter space.
 
         Parameters
         ----------
-        x : list
-            Value of the optimization variables in untransformed space.
+        x_independent : list
+            Value of the independent optimization variables in untransformed space.
 
         Returns
         -------
         list
             Optimization variables in transformed parameter space.
         """
-        x = np.array(x)
-        x_2d = np.array(x, ndmin=2)
+        x_independent = np.array(x_independent)
+        x_2d = np.array(x_independent, ndmin=2)
         transform = np.zeros(x_2d.shape)
 
         for i, ind in enumerate(x_2d):
@@ -2578,7 +2578,7 @@ class OptimizationProblem(Structure):
                 for value, var in zip(ind, self.independent_variables)
             ]
 
-        return transform.reshape(x.shape).tolist()
+        return transform.reshape(x_independent.shape).tolist()
 
     def untransform(self, x_transformed):
         """Untransform the optimization variables from transformed parameter space.
@@ -2716,7 +2716,7 @@ class OptimizationProblem(Structure):
             random: Any random valid point in the parameter space.
         seed : int, optional
             Seed to initialize random numbers. Only used if method == 'random'
-        burn_in: int, optional
+        burn_in : int, optional
             Number of samples that are created to ensure uniform sampling.
             The actual initial values are then drawn from this set.
             The default is 100000.
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -2837,11 +2837,7 @@ class OptimizationProblem(Structure):
 
             ind = self.get_dependent_values(ind)
 
-            if not self.check_bounds(ind):
-                continue
-            if not self.check_linear_constraints(ind):
-                continue
-            if not self.check_linear_equality_constraints(ind):
+            if not self.check_individual(ind, silent=True):
                 continue
 
             if not include_dependent_variables:
@@ -2953,6 +2949,39 @@ class OptimizationProblem(Structure):
 
             if not self.check_linear_constraints_dependency():
                 flag = False
+
+        return flag
+
+    @untransforms
+    @gets_dependent_values
+    def check_individual(self, x, silent=False):
+        """Check if individual is valid.
+
+        Parameters
+        ----------
+        x : list
+            Value of the optimization variables in untransformed space.
+
+        Returns
+        -------
+        bool
+            True if the individual is valid correctly, False otherwise.
+
+        """
+        flag = True
+
+        if not self.check_bounds(x):
+            if not silent:
+                warnings.warn("Individual does not satisfy bounds.")
+            flag = False
+        if not self.check_linear_constraints(x):
+            if not silent:
+                warnings.warn("Individual does not satisfy linear constraints.")
+            flag = False
+        if not self.check_linear_equality_constraints(x):
+            if not silent:
+                warnings.warn("Individual does not satisfy linear equality constraints.")
+            flag = False
 
         return flag
 
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -2691,6 +2691,13 @@ class OptimizationProblem(Structure):
             simplify=simplify,
         )
 
+        if self.n_linear_equality_constraints > 0:
+            problem = hopsy.add_equality_constraints(
+                problem,
+                self.Aeq,
+                self.beq
+            )
+
         return problem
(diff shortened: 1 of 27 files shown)


flo-schu:
## trying out new features

currently the fixtures for optimization problems are not on the `dev` branch. @schmoelder, maybe you can change this? 

Until then, there is a "fix". I am playing around with the surrogate model in the paper repository: https://github.com/flo-schu/bayesian-optimization
There I have just copy pasted the fixtures and work on the plots. If you clone this repo, you can try out the surrogate features

schmoelder:
> currently the fixtures for optimization problems are not on the `dev` branch. @schmoelder, maybe you can change this?

Yes, I will try to move it to `dev`.

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,119 @@
...
+            var_index = self.optimization_problem.get_variable_index(var.name)
+            if var.name in conditional_vars:
+                conditioning_value = conditional_vars[var.name]
+                var.lb = conditioning_value - eps
+                var.ub = conditioning_value + eps
+
+            else:
+                free_vars.update({var.name: var_index})
+
+        X, F = self.approximate_objectives(n_samples=n_samples)
```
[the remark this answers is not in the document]
flo-schu: yes. I'll change it.

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,119 @@
...
+            self.surrogate_model_M = gp_m
+
+        if CV is not None:
+            gp_cv = GaussianProcessClassifier()
+            gp_cv.fit(X, CV)
+            self.surrogate_model_CV = gp_cv
+
+
+
+    def estimate_objectives(self, X):
```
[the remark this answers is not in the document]
flo-schu: I agree. But I wasn't sure about the design of this anyway. Maybe it would be better to design the functions/properties of Meta scores, constraint violations, ... analogous to what you did in the optimization problem / results? And only replace `evaluate` with `estimate`?

CADETProcess/optimization/optimizationProblem.py
```diff
@@ -143,6 +143,10 @@ def wrapper(self, x, *args, get_dependent_values=False, **kwargs):
 
         return wrapper
 
+    def get_variable_index(self, variable):
```
schmoelder, line 146: please add a docstring.

CADETProcess/optimization/surrogate.py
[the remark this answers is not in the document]
schmoelder: Probably we should just mimic the `OptimizationProblem` interface for now. I.e. have individual functions for objectives, constraints, etc. This makes it easier to later fit it into existing architecture. 

Since at some point, I'd like to unify the `evaluate_` and evaluate_population` methods in the `OptimizationProblem` interface, I don't think we need to implement both here and already make the `evaluate` function work for both.

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,119 @@
...
+        X = self.optimization_problem.create_initial_values(
+            n_samples=n_samples,
+            method="random",
+        )
+        F = self.estimate_objectives(X)
+
+        return X, F
+
+
+    def condition_objectives(
```
schmoelder, line 94: Also add function to condition constraints etc.

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,119 @@
...
+    def _reset_bounds_on_variables(self):
+        for var, lb, ub in zip(
+            self.optimization_problem.variables,
+            self.lower_bounds_copy,
+            self.upper_bounds_copy
+        ):
+            var.lb = lb
+            var.ub = ub
+
+    def fit_gaussian_process(self, population: Population):
```
schmoelder, line 36: I'm wondering, maybe we should already pass a population in the constructor and provide an `update_model` function? This way we never encounter an un-initialized surrogate... But I guess it depends on where we put it in the architecture...

flo-schu:
I've made a very first approach for integration of Surrogate in `post_processing` and `OptimizationResults` it works for `U_NSGA` but it is still very WIPy.

Let's talk it through tomorrow and see which additional features we want and what can be improved, refactored, etc. :)

flo-schu:
not sure if the merge of dev into 42... was the right thing to do. Probably rebase would have been your preferred choice? If possible you can walk me through correcting this and doing it differently.

flo-schu:
okay, I have cleaned up `surrogate` module and removed it from `OptimizationResults` and `Optimizer`. Next week I will continue developing it for further optimization tasks and write some tests

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,381 @@
...
+        Parameters:
+        - X (np.ndarray): The input samples.
+
+        Returns:
+        - np.ndarray: The estimated meta scores.
+        """
+        M_est = self.surrogate_model_M.predict(X)
+        return M_est
+
+    def get_conditional_and_free_indices(self, conditional_vars={}):
```
schmoelder, line 140: Using mutable objects as default arguments is dangerous.

Better use 

```
def get_conditional_and_free_indices(self, conditional_vars=None):
    if conditional_vars = None:
        conditional_vars  = {}
```

Otherwise, all function calls will potentially refer to the same dict object.

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,381 @@
...
+)
+
+class Surrogate:
+    """
+    Surrogate class for optimization problems.
+    """
+
+    def __init__(
+        self,
+        optimization_problem: OptimizationProblem,
```
schmoelder, line 22: Only use OptimizationResults as parameter. The OptimizationProblem is also stored in that. This way, we can avoid inconsistencies.

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,381 @@
...
+        for v in self.optimization_problem.variable_names:
+            idx = self.optimization_problem.get_variable_index(v)
+            if v in conditional_vars:
+                cond_var_idx.append(idx)
+            else:
+                free_var_idx.append(idx)
+
+        return cond_var_idx, free_var_idx
+
+    def condition_constraints(self, conditional_vars={}):
```
schmoelder, line 164: see above...

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,381 @@
...
+        self,
+        optimization_problem: OptimizationProblem,
+        optimization_results: OptimizationResults,
+        n_samples=10000,
+    ):
+        """
+        Initialize the Surrogate class.
+
+        Parameters:
+        - optimization_problem (OptimizationProblem): The optimization problem.
```
schmoelder, line 30: Please use numpy style docstrings.

E.g.

```
def __init__(self, optimization_problem, population, n_samples):
    """
    Initialize the Surrogate class.

    Parameters
    ----------
    optimization_problem : OptimizationProblem
        The optimization problem.
    """
    [...]
```

CADETProcess/optimization/optimizationProblem.py
flo-schu: done

CADETProcess/optimization/surrogate.py
flo-schu: already changed

CADETProcess/optimization/surrogate.py
flo-schu: done

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,381 @@
...
+        A = self.optimization_problem.A.copy()
+        b = self.optimization_problem.b.copy().astype(float)
+
+        b_cond = b - A[:, cond_var_idx].dot(cvar_values)
+        A_cond = A[:, free_var_idx]
+
+        return A_cond, b_cond
+
+
+    def condition_optimization_problem(self, conditional_vars={}):
```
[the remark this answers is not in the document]
flo-schu: done

flo-schu:
wrapper around evaluate_objectives does not work because the optimization
problem can't deal with the new number of variabels

this has to do with `set_variables` and `get_dependent_variables` which return a lower number of variables than those that are actually present.

Therefore it is probably best to revert and use the old function (for now)
and find a way to constrain evaluation objects

Although this is messy it will probably easier yield a result than refactoring
condition optimization problem into something leaner and more proper 
whatever that means

flo-schu:
| :warning: GP instability for nonlinear constraints |
|----------------------------------------------------|

flo-schu:
While the behavior could be fixed, still the GP cannot be fitted on the problem. As suspected in:
> 2. standard scaling before fitting the GP. If the constraints/parameters operate on different scales, this might reduce the problem. This is unclear, if it will work, but it might improve the stability of the GP in any case. Then however, all X would need to be scaled before prediction (with a scaler provided to the surrogate, not to be confused with the transform method.

## other potential solutions:

- [ ] Scaling might also aid the problem, as F and G operate on highly different scales. Then standard scaling should be performed.
- [ ] Conceptually think about, whether it makes sense that nonlinear constraints can be violated in the same range where other individuals do not violate constraints.
- [ ] plot parameters / constraints predictions against true points

flo-schu:
> Scaling might also aid the problem, as F and G operate on highly different scales. Then standard scaling should be performed.

scaling solved the problem entirely

flo-schu:
branch `jos-problem` is now merged into this branch and tests are passing. Currently the surrogate model does not work (again) on the `no-waste` problem. This is because `estimate_x0` was removed from the module. I will integrate a reduced version in `find_x0.` With this, hopefully, the chebyshev computations and hopsy can be removed entirely from the `find_minimum` method.

- [ ] implement x0-finding algorithm at the edges of the problem in `find_x0`

When this is done, only the `condition_optimization_problem` needs to be refactored into a separate class, but this could also be done later

- [ ] refactor `condition_optimization_problem`

flo-schu:
x0 finding at the edges of the problem is increasingly difficult. I have decided to roll back on the state with a more costly iterative x0 estimation technique.

With some more time, it would definitely be possible to find a good solution for finding x0 at the boundaries of the problem. 

Ideally however the following algorithmus can be used:

1. analyze if the conditioned problem contains feasible space (e.g. with hopsy) (implemented but problematic due to hopsy issues)
2. if not continue with next conditioning slice
3. if feasible space exist, use existing evaluations to interpolate a new x0 (implemented)
4. if feasible space exists, but no points are there to interpolate, extrapolate into unexplored space. 

For 4. caveats exist:

- the surrogate in the unexplored space may return very unlikely predictions that are out of bounds for the true constraints or bounds of the problem
- in this case it may be better to not go into these regions by extrapolating and restrict the surrogate to the sampled range (which will most likely be close enough to the bounds)
- If the borders must be inside the problem, it may be a good idea to sample the corners of the problem in the first place. Because then also it will be much easier for the surrogate find good points near the boundary of a problem

flo-schu:
currently the module passes all tests and also is enabled to solve the no-waste problem. It could be merged. 

Open issues and improvements are tracked in https://github.com/fau-advanced-separations/CADET-Process/issues/52

schmoelder:
I think rebasing worked. However, there is now an issue with Python < 3.9:

E.g.
```
======================================================================
ERROR: test_surrogate (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_surrogate
Traceback (most recent call last):
  File "/usr/share/miniconda3/envs/test/lib/python3.9/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/share/miniconda3/envs/test/lib/python3.9/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/home/runner/work/CADET-Process/CADET-Process/tests/test_surrogate.py", line 6, in <module>
    from CADETProcess.optimization import (
  File "/home/runner/work/CADET-Process/CADET-Process/CADETProcess/__init__.py", line 34, in <module>
    from . import optimization
  File "/home/runner/work/CADET-Process/CADET-Process/CADETProcess/optimization/__init__.py", line 97, in <module>
    from .surrogate import Surrogate
  File "/home/runner/work/CADET-Process/CADET-Process/CADETProcess/optimization/surrogate.py", line 22, in <module>
    class Surrogate:
  File "/home/runner/work/CADET-Process/CADET-Process/CADETProcess/optimization/surrogate.py", line 162, in Surrogate
    def estimate_objectives(self, X):
TypeError: 'staticmethod' object is not callable

```
I'm not entirely sure why this suddenly comes up but it might be due to the recent changes in the Structure base class.

For reference: https://stackoverflow.com/questions/41921255/staticmethod-object-is-not-callable

This issue seems to e resolved in Python 3.10 but it's probably too early to "retire" 3.9. So we might have to look into this again.

PR #46, opened 2023-07-02
Fix linear constraints error in Scipy

flo-schu:
- [ ] checks if non-linear transforms are used in linear constraints problems
- [ ] transforms linear constraints to parameter space if linear


CADETProcess/transform.py
```diff
@@ -96,6 +96,14 @@ def __init__(
         self.allow_extended_input = allow_extended_input
         self.allow_extended_output = allow_extended_output
 
+
```
schmoelder, line 99: Maybe it's sufficient to later check `if isinstance(transform, NormLinearTransform)` since it's probably gonna be the only linear transform.

schmoelder:
>  checks if non-linear transforms are used in linear constraints problems

When implementing this, note that some optimizers (e.g. pymoo) don't need this check because they never work directly with the actual constraints matrix but call `check_linear_constraints` which untransforms automatically.

CADETProcess/transform.py
flo-schu: what about no transform? That's a linear transform mapping onto itself, I think.

flo-schu:
The implemented math did not pass the `LinearConstraintSooTestProblem` test problem. I've revisited the math in the transformation and realized that there needs to be some shifting done to `b` with respect to the constraint matrix `A`. Also scaling both `A` and `b` additionally by the outcome range of the target space allows mapping on target spaces that have not unit length. The true solution of the `TestProblem` can now be found, but I would prefer if someone with a more rigorous background than the two of us had a look at the problem. 

If you want, I'll walk you through my approach later. 
- Also I still need to implement those tests, if transformed samples from a constraint space match untransformed samples in an unconstrained space after back-transformation. 
- Also COBYLA breaks again for reasons I do not understand

CADETProcess/optimization/optimizationProblem.py
```diff
@@ -2187,7 +2187,7 @@ def evaluate_linear_constraints(self, x):
         """
         values = np.array(x, ndmin=1)
 
-        return self.A.dot(values) - self.b
+        return self.A_transformed.dot(values) - self.b_transformed
```
schmoelder, line 2190: Here, I think we need to be careful. This function should also work with the `untransform=True` flag which creates x from x_untransformed and consequently evaluates `A`

tests/optimization_problem_fixtures.py
```diff
@@ -110,8 +110,22 @@ def test_if_solved(self, optimizer, decimal=7):
...
+
+        self.add_variable('var_0', lb=-2, ub=2, transform="linear")
+        self.add_variable('var_1', lb=-2, ub=2, transform="linear")
+        self.add_linear_constraint(['var_0', 'var_1'], [-1, -0.5], 0)
+
+
+
 class LinearConstrainedMooTestProblem(TestProblem):
-    """Function curtesy of Florian Schnunk and Samuel Leweke."""
+    """Function curtesy of Florian Schnunck and Samuel Leweke."""
```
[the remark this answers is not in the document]
flo-schu: Hahaha ich hab's auch noch mal falsch geschrieben

schmoelder:
Tests are failing because scipy==1.11 requires Python>=3.9
For CADET-Process 0.8, I wanted to test for [3.9, 3.10, 3.11] anyways so I think it's fine to update setup.cfg and pipeline.yml

flo-schu:
- [x] @flo-schu transform non-linear transforms to nonlinear constraints (maybe, if possible...) --> converted to #48 
- [x] @flo-schu create reproducible example for COBYLA error --> commented in #47

flo-schu:
> What we still need to think about is how to handle the linear equality constraints. I don't know if we can use the same transformations as for the inequality constraints.

I'd assume, yes. However, if I remember correctly, COBYLA does not allow equality constraints (any longer?)

```py
    for ic, con in enumerate(constraints):
        # check type
        try:
            ctype = con['type'].lower()
        except KeyError as e:
            raise KeyError('Constraint %d has no type defined.' % ic) from e
        except TypeError as e:
            raise TypeError('Constraints must be defined using a '
                            'dictionary.') from e
        except AttributeError as e:
            raise TypeError("Constraint's type must be a string.") from e
        else:
            if ctype != 'ineq':
                raise ValueError("Constraints of type '%s' not handled by "
                                 "COBYLA." % con['type'])
```

schmoelder:
> I'd assume, yes. However, if I remember correctly, COBYLA does not allow equality constraints (any longer?)

That's possible but this could be handled when checking the compatibility of Optimizer/OptimizationProblem (see https://github.com/fau-advanced-separations/CADET-Process/blob/29eb5cc463892b4e969dcd07f3cd9fb968bcce23/CADETProcess/optimization/optimizer.py#L278)

CADETProcess/optimization/optimizationProblem.py
```diff
@@ -2541,20 +2678,91 @@ def parameters(self):
...
         if self.n_variables == 0:
             flag = False
         if self.n_objectives == 0:
             flag = False
+        if self.n_linear_constraints > 0 and not ignore_linear_constraints:
+            flag_ = self.check_linear_constraints_transforms()
+            flag = flag and flag_
+
+            flag_ = self.check_linear_constraints_dependency()
+            flag = flag and flag_
```
flo-schu, line 2765: Good spot! This would have been annoying ;)

flo-schu:
## linear equality constraints

comments to commit https://github.com/fau-advanced-separations/CADET-Process/pull/46/commits/388f873a99921feb3defd243b060acbe5c0a3ee3
trying to attack this problem with brute force (rejection sampling) is expensive.
With a tolerance of `1e-3` in a uniformly samples space in a 3-variable problem with 1 linear equality constraint, 5 in 100000 samples satisfy the constraints. This is problematic for the `create_initial_values` function.

Maybe a better approach would be to specify the equality constraint as a dependent variable problem, and a new bound for the independent variable.


### COBYLA

does not allow constraints of the type "eq"

suggestion, don't allow equality constraints for `COBYLA`

### TrustConstr

warns that function may be linear (which it is for `LinearEqualityConstraintsSooTestProblem`) and suggest to set the Hessian to zero.

when the matrix is linear transformed, Jacobian is singular. Cannot solve problem

suggestion, don't allow equality constraints for `TrustConstr`

### SLSQP 

works without transform and with transform very quickly. For me the method of choice if linear transforms are present

### U_NSGA3

dimensionality problem when equality constraints are set. Succeeds only when the burn is is large or when the tolerance of the constraints is relatively high (which will result in inaccurate solutions).

Suggestion, don't allow equality constraints as long as `hopsy` does not handle equality constraints.

PR #57, opened 2023-09-01
Fix/update docs build instructions

ronald-jaepel:
Short addition of detail to the doc building instructions


schmoelder:
I updated the section on `dev`.

PR #58, opened 2023-09-01
Add save_to_csv functionality to simulationResults

ronald-jaepel:
# ToDos:

- [ ] Add test


schmoelder:
I think it might be a good idea to implement the functionality that saves the `csv` on the `Solution` layer. Then, we could still call that from `SimulationResults` when iterating over the units and recorded solutions.

PR #63, opened 2023-10-12
Fix missing prop_cycler error

ronald-jaepel:
Found a second spot in solution.py that referenced `ax._get_lines.prop_cycler` and replaced it with `plt.rcParams["axes.prop_cycle"].by_key()["color"]`


schmoelder:
I ammended this to `dev`

PR #73, opened 2023-11-28
Add ports / MCT Model

### LINKED ISSUE #3: Add 2D-GRM
In **CADET-Core**, a [2D General Rate Model](https://cadet.github.io/master/interface/unit_operations/2d_general_rate_model.html) is available. Moreover, other 2D units are being developed, such as [MCT](https://github.com/modsim/CADET/pull/95). To support these in CADET-Process, ports need to be added to the `FlowSheet`. 

This is, however, not trivial since flow rates are treated differently in CADET-Process than they are in CADET-Core. While in CADET-Core, all flows between units are always explicitly specified, in CADET-Process they are automatically determined in a preprocessing routine. Here, the concept of `output_state` is introduced which specifies the ratio of all outgoing flows. Consider the following `FlowSheet`:

```mermaid
graph LR
    A[Inlet] -->B[Column]
    B -->|0.5| C[Outlet A]
    B -->|0.5| D[Outlet B]
```

The flow entering the column is split 50/50 to Outlets A and B, respectively.

A 2D unit has multiple inlet ports. Here there are several options:
- **Option 1**: Expose the ports on `FlowSheet` level s.t. users can explicitly define flow from certain upstream units to the corresponding port and then from the outlet port to specific downstream units.
- **Option 2**: Treat the inlet as a single port which distributes all incoming flow according to some internal `input_state` which is recombined at the column outlet before being sent to downstream units according to the unit's `output_state`

Option 1 is more flexible but also more complicated to setup. Also, what's the notion of an `output_state` here? Should there be an output state per port?

Option 2 is simpler but also less flexible.

- [ ]  Handling of nrad in CadetAdapter
- [ ]  Translation between nrad and nports
- [ ]  Adding nrad to create_LWE
- [ ]  Tests

### LINKED ISSUE #70: Add support for parameter multiplexing
For many parameters in CADET-Core, the dimensions can change depending on whether the parameter is assumed to be dependent (or independent) on components, sections, particle types, etc.

E.g., consider the `FILM_DIFFUSION` of the [GRM](https://cadet.github.io/master/interface/unit_operations/general_rate_model.html). Here, the parameter can be specified per
- component
- section
- particle

To avoid ambiguities, a multiplex-parameter is introduced. In this example, `FILM_DIFFUSION_MULTIPLEX` determines whether the parameter is treated as component-, type-, and/or section-independent. This field is optional. When left out, multiplexing behavior is inferred from the length of `FILM_DIFFUSION` (if possible). Valid modes are:
- 0: Component-dependent, type-independent, section-independent; length of  `FILM_DIFFUSION` is `NCOMP`
- 1: Component-dependent, type-independent, section-dependent; length of length of  `FILM_DIFFUSION` is `NCOMP` * `NSEC`; ordering is section-major
- 2: Component-dependent, type-dependent, section-independent; length of  `FILM_DIFFUSION` is `NCOMP` * `NPARTYPE`; ordering is type-major
- 3: Component-dependent, type-dependent, section-dependent; length of  `FILM_DIFFUSION` is `NCOMP` * `NPARTYPE` * `NSEC`; ordering is section-type-major

However, it's not possible to directly mirror this behaviour in CADET-Process since sections are handled differently (using `Events`). Moreover, some features might still be missing, such as multiple particle types.

For this reason, I propose that we add flags (with reasonable defaults) which are then later used to infer the multiplex value generating the config for CADET-Core. This will also allow us to extend the functionality step by step.

In practice, this would look something like this:

```Python
class GRM(UnitBaseClass):
    axial_dispersion = MultiplexedParameter(modes=(axial_dispersion_is_component_dependent, axial_dispersion_is_component_dependent, axial_dispersion_is_component_dependent), default=0)
    axial_dispersion_is_component_dependent = Const(value=True)
    # axial_dispersion_is_section_dependent = Const(value=True) # This one we do not really need since they are always section dependent in CADET-Process
    axial_dispersion_is_particle_type_dependent = Bool(default=False) # Note, multiple particle types are not yet implemented
```

The `MultiplexedParameter` would be responsible of checking the parameter dimension on setting (and getting) by reading the values of the flags.

TODOs:
- [ ] Allow differing flow directions for different channels / radial cells in 2D-models

schmoelder:
This PR adds ports to UnitOperations in CADET process which are required by models such as the 2D-GRM.

For starters, we will test this with the Multi-Channel Transport-Model

To fix #3, we also need to fix #70.

## To do
- [x] Add ports to `add_connection` method
- [x] Add discretization class for MCT
- [x] Add SolutionRecorder for MCT
- [x] Store connections with ports in `FlowSheet`
    - [x] Add tests
    - [x] How to expose discretization n_rad to flow_sheet/ports
- [ ] Add check methods
    - [ ] Unconnected ports
    - [ ] Changes in `discretization` configuration
    - [ ] flow rate balance
- [ ] Open question: Should ports have individual output states?
- [x] Add parameters to the Name Converter for the MCT in cadetAdapter
- [ ] Create a SizedMatrix Type and/or check SizednDArray if it meets our requirements
- [x] Add MCT tests in test_unit_operation
     - [x] add create_MCT
- [ ] In unitOperation change flow_direction into array of Switches
- [ ] Check dimensionality for all unit operations, see also in solution

## Open question
Should the MCT (and other potential 2D models) inherit from `TubularReactor`? How would this look like.
Should every channel of the MCT have a port by default?


schmoelder:
Ok, I can now setup an `MCT` unit operation. We should still test whether all parameters are transferred correctly to CADET in the CadetAdapter.

Next steps should address the ports. Please add some tests so that we have something to work with. Also, we should work on adding the multiplex features (see #70).

hannahlanzrath:
Test cases for the MCT

image

tests/test_unit_operation.py
```diff
@@ -60,6 +64,18 @@ def create_tubular_reactor(self):
 
         return tube
 
+    def create_MCT(self):
+        mct = MCT(ComponentSystem(1), nchannel=3, name='test')
+
+        mct.length = length
```
schmoelder, line 70: We should consider moving this outside the `Test_Unit_Operation` class definition s.t. we could potentially reuse the same fixture for other tests.

schmoelder:
Sorry, merged the wrong branch. :sweat:

PR #76, opened 2023-12-01
Fix bad initial value for Yamamoto method

ronald-jaepel:
Change initial value from the automatic middle of the boundary space (500 & 500) to (1, 1), as an initial value of 500 can lead to numerical errors.

Also adapted yamamoto.py to PEP8.

### DIFF
--- a/CADETProcess/tools/yamamoto.py
+++ b/CADETProcess/tools/yamamoto.py
@@ -255,7 +255,7 @@ def _fit_yamamoto(log_c_salt_at_max_M, log_gradient_slope, lambda_):
         return yamamoto_equation(c_s, lambda_, nu, k_eq)
 
     results, pcov = curve_fit(
-        yamamoto_wrapper, log_c_salt_at_max_M, log_gradient_slope, bounds=bounds
+        yamamoto_wrapper, log_c_salt_at_max_M, log_gradient_slope, bounds=bounds, p0=(1, 1)
     )
 
     return results


CADETProcess/tools/yamamoto.py
```diff
@@ -254,8 +253,7 @@ def _fit_yamamoto(log_c_salt_at_max_M, log_gradient_slope, lambda_):
     def yamamoto_wrapper(c_s, nu, k_eq):
         return yamamoto_equation(c_s, lambda_, nu, k_eq)
 
-    results, pcov = curve_fit(
-        yamamoto_wrapper, log_c_salt_at_max_M, log_gradient_slope, bounds=bounds
-    )
```
schmoelder, line 259: tbh, I like this style better.

schmoelder (changes requested):
Thanks for taking care of this!

Please use separate commits for formatting and fixing issues.
In terms of style, I agree, we should probably update the entire code base with a linter such as `ruff`. There are, however some cases where I'd like to not use the default settings. E.g., I really dislike the way it uses linebreaks with parenthesis (see separate comment).

CADETProcess/tools/yamamoto.py
ronald-jaepel: Alright.

schmoelder:
Oh, and please use the "rebase and squash" option.

Issue #75, opened 2023-12-01
Computation time increased ~100 times in v0.8 for 20 columns SMB system


dion-is:
After upgrading to v0.8 calculation time for 20 columns carousel system increased from 5  min to 10 hours. The case is attached.

[test_20_columns.zip](https://github.com/fau-advanced-separations/CADET-Process/files/13527371/test_20_columns.zip)

dion-is:
@schmoelder , yes,  this is just the time for pre-processing. 

I believe it is not related to solving flow rates, I'm using my own function to do it. I ran a profiler and it spends most of the time in event sorting.

schmoelder:
We created a PR (#77) where we're trying out some things to improve this. On my local machine, your code now runs in 95 s. This is still not great since the simulation itself only takes 5 s to run but I hope it makes it at least usable again.

Note that this is still mostly untested so I hope this does not break anything. Please also let us know if you have further ideas on how to improve this situation while also keeping the flexibility that CADET-Process generally provides.

Issue #67, opened 2023-10-30
Improving flowrate calculation


daklauss:
As described in [Forum](https://forum.cadet-web.de/t/improving-the-flowrate-calculation/795) we found a possible improvement to the current simpy solution to calculate the flow rates in `flowSheet.py`.

## current solution

There are two functions to do the work. The main function `get_flow_rates()` and a function to solve: `solve_flow_rates()`. 

The main function returns a dictonary consisting of key/value pairs where the keys are the unit-names and the values themselves are dictionaries consisting destinations, total_out and/or total_in.

## goal

The goal is to optimize the solve function by setting up a matrix and solve with numpy instead of using the symbolic solution with simpy. Depending on how complicated this will be to implement it may be possible that `solve_flow_rates()` as a function will become redundand.

schmoelder:
Unfortunately, there are now some issues with the `CarouselBuilder`.

```
# Component System
from CADETProcess.processModel import ComponentSystem
component_system = ComponentSystem(2)

# Binding Model
from CADETProcess.processModel import Linear
binding_model = Linear(component_system)
binding_model.adsorption_rate = [6, 8]
binding_model.desorption_rate = [1, 1]

from CADETProcess.processModel import LumpedRateModelWithoutPores
column = LumpedRateModelWithoutPores(component_system, name='master_column')
column.length = 0.6
column.diameter = 0.024
column.axial_dispersion = 4.7e-7
column.total_porosity = 0.7

from CADETProcess.processModel import Inlet, Outlet
i_1 = Inlet(component_system, name='i_1')
i_1.c = [0, 0]
i_1.flow_rate = 60/60/1e6

i_2 = Inlet(component_system, name='i_2')
i_2.c = [10, 10]
i_2.flow_rate = 30/60/1e6

o_1 = Outlet(component_system, name='o_1')
o_2 = Outlet(component_system, name='o_2')

from CADETProcess.modelBuilder import SerialZone, ParallelZone

z_serial = SerialZone(component_system, 'z_serial', 3)
z_parallel = ParallelZone(component_system, 'z_parallel', 3)

from CADETProcess.modelBuilder import CarouselBuilder


# %% Works:

# Inlet -> SerialZone -> Outlet
builder = CarouselBuilder(component_system, 'multi_zone')
builder.column = column

builder.add_unit(i_1)
builder.add_unit(o_1)

builder.add_unit(z_serial)

builder.add_connection(i_1, z_serial)
[...]

PR #77, opened 2023-12-05
Improve preprocessing time

schmoelder:
As mentioned in #75, pre-processing can take a long time, especially for complex systems with many units and parameters. In this PR, we try to address some of the issues.

## Options
- [x] Improve flow rate calculations (see #67/#71)
- [x] Cache more properties
- [x] Avoid unnecessary checks (e66ec1b)
- [ ] From previous experience: locked properties can cause side effects. Need more thorough tests & thoughts
- [ ] Need feedback from Dennis

Consider further profiling to identify current bottlenecks and consider architecture changes that further speed this up

### DIFF
--- a/CADETProcess/modelBuilder/carouselBuilder.py
+++ b/CADETProcess/modelBuilder/carouselBuilder.py
@@ -140,12 +140,9 @@ class CarouselBuilder(Structure):
 
                 flow_sheet.add_connection(origin, destination)
 
-        flow_rates = self.flow_sheet.get_flow_rates()
         for zone in self.zones:
             output_state = self.flow_sheet.output_states[zone]
             flow_sheet.set_output_state(zone.outlet_unit, output_state)
-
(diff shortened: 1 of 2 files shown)


CADETProcess/modelBuilder/carouselBuilder.py
```diff
@@ -128,13 +128,10 @@ def add_inter_zone_connections(self, flow_sheet):
 
                 flow_sheet.add_connection(origin, destination)
 
-        flow_rates = self.flow_sheet.get_flow_rates()
         for zone in self.zones:
```
schmoelder, line 131: I wonder if this is even necessary here. I think that output states are entirely handled by `Events`.

schmoelder:
@dion-is, happy new year! :)

Did you ever get a chance to test this PR?

dion-is:
@schmoelder , sorry, I just rolled back to the previous version for now. Will test this later.

PR #79, opened 2023-12-07
Fix behavior of comparisons to references with multiple components

ronald-jaepel:
As discussed [here](https://forum.cadet-web.de/t/bug-in-class-referenceio/817/4), this PR fixes previously unhandled cases with multi-component references.


schmoelder:
LGTM.

But before merging, let's have a short chat regarding testing, focusing on the following aspects:
- general status of testing references/difference metrics
- reference_single vs reference (is reference_single still required?)
- what about component systems with sub-species?

PR #80, opened 2023-12-07
Add examples for binding parameters

ronald-jaepel:
Here I've expanded the example library to include SMA parameter determinations using the Yamamoto method and the inverse method.

This contains changes from the branch #66  and should be merged after #66  is merged.

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -23,6 +23,8 @@ examples/batch_elution/process
 examples/load_wash_elute/index
 examples/recycling/index
 examples/characterize_chromatographic_system/column_transport_parameters
+examples/characterize_chromatographic_system/binding_model_parameters
+examples/characterize_chromatographic_system/Yamamoto_method
 ```
 
 ```{toctree}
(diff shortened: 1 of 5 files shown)


schmoelder:
Thanks for this! #66 is now merged; please rebase onto `dev`.

This could be a good idea to reconsider the structure of our examples. A long standing wish is to separate them into individual repositories where we can side-load optimization results using something like CADET-RDM.

schmoelder:
You might also wanna include the example in the docs [here](https://github.com/fau-advanced-separations/CADET-Process/blob/d497d0c6f23c9601e0efc8e6985bde240d129147/docs/source/index.md?plain=1#L25)

ronald-jaepel:
Done.

ronald-jaepel:
The test fails due to a disk-cache error that I can not reproduce locally. On Windows and Linux all tests complete fine.

ronald-jaepel:
Tests completed, todos done. I think this can be merged.

schmoelder:
Nice, just for peace of mind, did you also build the docs locally to check that they work properly?
(might also be a good idea for further CI tests).

PR #91, opened 2024-02-01
Improve behavior of evaluate_callbacks with individuals without x_untransformed.

ronald-jaepel:
Improve behavior of OptimizationProblem.evaluate_callbacks when it gets handed an Individual without x_untransformed.


schmoelder:
@ronald-jaepel I changed the approach and added a method to automatically create new individuals from the OptimizationProblem (given some `x`). As mentioned in the forum, eventually, I would also like to use that method for wrapping all the evaluation methods s.t. it is no longer required to pass individuals at all. Moreover, I’d like to also add a `create_population` method.

PR #102, opened 2024-03-02
 Change behavior for adding Difference Metrics

### LINKED ISSUE #36: add section on how to implement your own difference metric
- needs to subclass DifferenceBase
- implement _evaluate function that calculates the metric
- change metric input to class instead of string

from
```python
comparator.add_difference_metric(
    'PeakHeight', reference, 'outlet.outlet',
)
```

```python
from CADETProcess.comparison.difference import PeakHeight
comparator.add_difference_metric(
    PeakHeight, reference, 'outlet.outlet',
)
```

to create own metric

```python
from CADETProcess.comparison.difference import DifferenceBase

class MyMetric(DifferenceBase):
    def _evaluate(self, ...)
        ...

comparator.add_difference_metric(
    MyMetric, reference, 'outlet.outlet',
)
```

schmoelder:
This PR aims at restructuring the behavior for adding difference metrics to the `Comparator` class.

Closes #36 

In practice, this could would then look something like 
```
from CADETProcess.processModel import ComponentSystem
from CADETProcess.comparison import Comparator
from CADETProcess.comparison import AbsoluteArea
from tests.test_solution import TestSolutionIOGaussian

component_system = ComponentSystem(2)
tracer_peak = TestSolutionIOGaussian(component_system)

comparator = Comparator()

metric = AbsoluteArea(tracer_peak)
comparator.add_difference_metric(metric, 'outlet.outlet')
```

Note, it's not required any more to register the `reference` with the `Comparator` since this was only required for instantiating the difference metrics. This is now handled outside.

Since this would be a breaking change, I tried to add deprecation warnings s.t. we can provide backwards compatibility for some time.

## To do
- [x] Discuss design changes with @ronald-jaepel 
- [x] Add tests for `Comparator`
- [x] Make a plan for deprecating such interfaces
- [x] Update documentation

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -5,49 +5,77 @@ on:
     branches:
       - master
       - dev
+      - test_ci
   pull_request:
 
 jobs:
   test-job:
     runs-on: ${{ matrix.os }}
+
     defaults:
       run:
         shell: bash -l {0}
+
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
+        python-version: ["3.10", "3.11", "3.12"]
+        include:
+          - os: windows-latest
+            python-version: "3.12"
+          - os: macos-latest
+            python-version: "3.12"
+
+    env:
+      CONDA_FILE: environment.yml
+
     steps:
-    - uses: actions/checkout@v3
-    - name: Cache conda
-      uses: actions/cache@v3
-      env:
-        # Increase this value to reset cache if etc/example-environment.yml has not changed
-        CACHE_NUMBER: 2
-      with:
-        path: ~/conda_pkgs_dir
-        key:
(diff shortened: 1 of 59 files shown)


hannahlanzrath (approved):
Summary in my own words, so I can make sure I understood everything correctly: 
- `add_reference `was deprecated but is backwards compatible until v.1.0
- instead of directly passing a reference (e.g. experimental chromatogramm) to the comparator and then passing difference metric as an argument,
the difference metric is now given as an object to the comparator
- This is making use of the already existing structures of the `DifferenceBase` class
- This takes away one layer of abstraction, instead of choosing from a list of strings pointing to objects,
the user directly defines the object

--> Less nested, more Pythonic, and it makes it easier to extend difference metrics in the future

I think this is a good design decision and good to go from my side!

schmoelder:
Thanks for your feedback!

yes, that's correct. Also, this should make adding custom user metrics much easier: inherit from base and implement the interface. no need to register in the module itself.

PR #111, opened 2024-03-14
Add IPOPT Optimizer

schmoelder:
This PR adds support for interior point optimzer IPOPT using [cyipopt](https://github.com/mechmotum/cyipopt).

@flo-schu I guess this still needs some work but it's been in a local branch for a long time and maybe it would be a good opportunity to include it in the test runners and maybe even for the optimization study.

Could you please jog my memory and briefly tell me what is required to run the `test_optimizer_behaviour` using pytest?

### DIFF
--- a/CADETProcess/comparison/comparator.py
+++ b/CADETProcess/comparison/comparator.py
@@ -84,7 +84,7 @@ class Comparator(Structure):
         return self._metrics
 
     @property
-    def n_diffference_metrics(self):
+    def n_difference_metrics(self):
         """int: Number of difference metrics in the Comparator."""
         return len(self.metrics)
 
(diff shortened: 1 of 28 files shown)


flo-schu:
The new optimizer needs to be inserted here:

```python
@pytest.fixture(params=[
    SLSQP,
    TrustConstr,
    U_NSGA3,
    GPEI,
    NEHVI,
])
def optimizer(request):
    return request.param()
```

that's it. Then call the test with `pytest tests/test_optimizer_behavior.py::test_from_initial_values`

By the way, in `dev` currently both, `test_from_initial_values` and `test_convergence` are used. Both tests are almost the same and I don't think the `test_convergence` test is necessary.

schmoelder:
@hannahlanzrath, this is another optimizer that uses gradient information. This is WIP, but maybe you'd be up for testing it a bit, especially when considering how Jacobians (and Hesse) matrices computed?

schmoelder:
@hannahlanzrath this is the optimizer we discussed with Anton earlier this week. I just realized that somehow you were already assigned. Is this something you would like to explore? I think it's not too complicated and might be a good challenge to better get to know the optimization module in CADET-Process. Obviously, this can wait until after your thesis is handed in.

hannahlanzrath:
Yeah, Looks interesting. Maybe something that can also be combined with the inclusion of parameter sensitivities.

PR #114, opened 2024-03-17
Improve optimization post processing

### LINKED ISSUE #99: Overhaul post processing in Optimization
- Unify post processing (generation vs individual)
- Always run post processing for last generation
- Speed up post_processing (e.g. allow parallel evaluation for callbacks)
- Store system information in OptimizationResults (and export them to file).

schmoelder:
Closes #99.

## To do
- [x] Unify post_generation/post_evaluation_processing
- [x] Improve cache pruning: Keep members of pareto front for future evaluations (e.g. callbacks)
- [/] Evaluate callbacks in parallel (was already working)
- [x] Always run post_processing / callbacks for final generation (relevant when evaluation frequencies != 1)
- [x] Store optimization results, including run and system information after termination (use checkpoint as template)

---

In future, I might want to revisit caching. I'm still not happy with how it's handled, i.e. which object "holds" the cache. Right now, it's the OptimizationProblem. However, it might also be a good idea to just optionally pass in a cache object on evaluation call.

### DIFF
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -112,7 +112,6 @@ class OptimizationProblem(Structure):
         self._evaluation_objects_dict = {}
         self._evaluators = []
 
-        self.cached_evaluators = []
         self.use_diskcache = use_diskcache
         self.cache_directory = cache_directory
         self.setup_cache()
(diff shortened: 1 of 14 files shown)


CADETProcess/optimization/optimizationProblem.py
flo-schu, line 1: I can't really judge what the caching-changes do. So I leave it unreviewed.

CADETProcess/optimization/optimizer.py
flo-schu, line 1: This look **much cleaner** than before :clap: I have not checked all the details, but the implementation looks solid.

flo-schu (commented):
I think this is great work. I have left the files that I don't understand (caching) unreviewed. There is only one minor proposal for changing the implementation

ronald-jaepel:
Tested on george's script and the callbacks are indeed calculated at the given frequency as well as after the final evaluation.

PR #116, opened 2024-03-25
Fix/non linear variables uninvolved in linear rules

ronald-jaepel:
As mentioned in issue #117

### DIFF
--- a/tests/optimization_problem_fixtures.py
+++ b/tests/optimization_problem_fixtures.py
@@ -165,29 +165,30 @@ class LinearConstraintsSooTestProblem(TestProblem):
     def setup_variables(self, transform):
         self.add_variable('var_0', lb=-2, ub=2, transform=transform)
         self.add_variable('var_1', lb=-2, ub=2, transform=transform)
-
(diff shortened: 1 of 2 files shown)


tests/optimization_problem_fixtures.py
```diff
@@ -165,29 +165,30 @@ def __init__(self, transform=None, has_evaluator=False, *args, **kwargs):
     def setup_variables(self, transform):
         self.add_variable('var_0', lb=-2, ub=2, transform=transform)
         self.add_variable('var_1', lb=-2, ub=2, transform=transform)
-
+        self.add_variable('var_2', lb=0, ub=2, transform="log")
```
schmoelder, line 168: If I understood correctly, you're extending the Rosenbrock Problem, right?

What I would suggest instead is making the third variable optional s.t. we still have the pure/original Rosenbrock problem which we might be useful when trying to benchmark against other packages etc. 

Alternatively, we could also simply add a new TestCase/Problem (that might inherit from Rosenbrock) with other transforms and/or variable dependencies (something that's generally still missing in the fixtures).

Issue #104, opened 2024-03-07
Remove peak height from shape function


ronald-jaepel:
Edit:

We added a Deprecation warning in #119. In the next release, we can remove this entirely.

ronald-jaepel:
Solved in #119

PR #119, opened 2024-03-26
Remove height from shape metric

ronald-jaepel:
solves #104

### DIFF
--- a/CADETProcess/comparison/difference.py
+++ b/CADETProcess/comparison/difference.py
@@ -466,6 +466,7 @@ class Shape(DifferenceBase):
     def __init__(
             self, *args,
             use_derivative=True, normalize_metrics=True, normalization_factor=None,
+            include_height=True,
             **kwargs):
         """Initialize Shape metric.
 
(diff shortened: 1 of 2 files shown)


CADETProcess/comparison/difference.py
```diff
@@ -466,6 +466,7 @@ class Shape(DifferenceBase):
     def __init__(
             self, *args,
             use_derivative=True, normalize_metrics=True, normalization_factor=None,
+            include_height=False,
```
schmoelder, line 469: Please udpate the docstring

CADETProcess/comparison/difference.py
```diff
@@ -535,14 +539,21 @@ def __init__(
 
     @property
     def n_metrics(self):
+        n_metrics = 2
+
+        if self.include_height:
+            n_metrics += 1
+
         if self.use_derivative:
```
schmoelder, line 547: Should we also only optionally compute peak height difference for the derivatives (i.e. if self.include_height)?

schmoelder:
> solves #104

You can also link to Issues to close them automatically when adding them in the development panel on the right:

grafik

CADETProcess/comparison/difference.py
ronald-jaepel: I've changed it to do that.
```
        include_height : bool, optional
            If True, also return a metric for the height difference.
            If `use_derivative` is also true, also return height differences
            for the derivatives.
```

ronald-jaepel:
Done with requested changes. Ready to merge imo.

PR #122, opened 2024-04-05
Extend example of binding parameter estimation to include parameter transformation

ronald-jaepel:


ronald-jaepel:
I've extended the documentation to also include a section on how to use variable dependencies with kEq and kKin. IMO this PR is ready for Review and then merging.

docs/source/user_guide/optimization/variable_dependencies.md
```diff
@@ -61,3 +58,67 @@ optimization_problem.add_variable_dependency('var_2', ['var_0', 'var_1'], transf
 ```
 
 Note that generally bounds and linear constraints can still be specified independently for all variables.
+
+## Adsorption rates example
+
+For instance, consider an adsorption proces with an adsorption rate $k_a$ and a desorption rate $k_d$. Both
```
schmoelder, line 64: Please use one line per sentence.

examples/characterize_chromatographic_system/binding_model_parameters.md
```diff
@@ -264,7 +300,12 @@ if __name__ == '__main__':
         )
 
 
-    optimization_problem.add_callback(callback, requires=[simulator])
+    optimization_problem.add_callback(callback, requires=[simulator], callbacks_dir="tmp")
```
schmoelder, line 303: is the callbacks_dir required here?

schmoelder (changes requested):
Thanks for your input. I have just some minor comments. Otherwise, LGTM! :)

examples/characterize_chromatographic_system/binding_model_parameters.md
ronald-jaepel: Without that, the optimization_problem.evaluate_callbacks(ind) in line 308 throws an error.

PR #125, opened 2024-04-15
Use seconds in plot time axis

schmoelder:
This PR adds an option to change the time unit in plots to "seconds". Currently, minutes are always used as a unit in plots. However, it would be nice to make this behavior configurable since everything is usually assumed to be in base SI-Units.

## Todo
- [x] Check other plot methods and implement where necessary.
- [x] Is there other behavior we might want to change?

### DIFF
--- a/CADETProcess/comparison/comparator.py
+++ b/CADETProcess/comparison/comparator.py
@@ -4,9 +4,11 @@ import functools
 
 import numpy as np
 import matplotlib.pyplot as plt
+from matplotlib.figure import Figure
+from matplotlib.axes import Axes
 
 from CADETProcess import CADETProcessError
-from CADETProcess import plotting
+from CADETProcess import plotting, SimulationResults
 from CADETProcess.dataStructure import Structure, String
 from CADETProcess.dataStructure import get_nested_value
 from CADETProcess.solution import SolutionBase
@@ -223,7 +225,9 @@ class Comparator(Structure):
 
     __call__ = evaluate
 
-    def setup_comparison_figure(self, plot_individual=False):
+    def setup_comparison_figure(
+        self, plot_individual: bool = False
+    ) -> tuple[list[Figure], list[Axes]]:
         """Set up a figure for comparing simulation results.
 
         Parameters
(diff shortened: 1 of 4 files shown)


schmoelder (commented):
Thanks @AntoniaBerger for helping with this PR. Since I recently started adding type annotations on whatever I work on, please also take the opportunity to add them here. If you're super motivated, you could also add them for all the modules you've "touched", but that's optional.

environment.yml
schmoelder, line 1: Please do not delete the environment! :sweat_smile:

CADETProcess/comparison/comparator.py
```diff
@@ -267,15 +272,16 @@ def setup_comparison_figure(self, plot_individual=False):
...
-        simulation_results,
-        axs=None,
-        figs=None,
-        file_name=None,
-        show=True,
-        plot_individual=False,
-        use_minutes=True,
-    ):
-        """Plot the comparison of the simulation results with the reference data.
+        simulation_results: list[SimulationResults],
```
schmoelder, line 275: I know, I wasn't always super consistent myself, but please use the following indentation style:
```
def foo(
        arg_1: bool False,
        ) -> NoReturn:
    pass
```

Please also check the other methods.

CADETProcess/comparison/comparator.py
```diff
@@ -1,12 +1,15 @@
 import copy
 import importlib
 import functools
+from typing import Any
```
schmoelder, line 4: If you don't use `Any`, please remove from list of imports. (Also see below)

CADETProcess/comparison/comparator.py
```diff
@@ -284,7 +297,8 @@ def plot_comparison(
             If True, displays the figure(s) on the screen.
         plot_individual : bool, optional
             If True, generates a separate figure for each metric.
-
+        use_minutes: bool, optional
+            Option to use x-aches (time) in minutes, default is set to True.
         Returns
```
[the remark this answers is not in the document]
schmoelder: (also see below)

Issue #121, opened 2024-03-29
optimization_problem.evaluate_callbacks(ind) doesn't evaluate if callback frequency is set to != 1


ronald-jaepel:
When using 
```
    optimization_problem.add_callback(callback, requires=[simulator], frequency=5)
    x = [215, 1782, 61, 41, 0.0073, 0.00367, 3, 5, 2, 5]
    ind = optimization_problem.create_individual(x)
    optimization_problem.evaluate_callbacks(ind)
```
no callbacks are evaluated because during `.evaluate_callback` the check 
```
            if not (
                    current_iteration == 'final'
                    or
                    current_iteration % callback.frequency == 0):
                continue
```
causes it to skip all callbacks, as it's neither the "final" iteration, nor is the current_iteration divisible by 5.

schmoelder:
Fixed in b38e443aa8761ca91f0d24481719442445726cd4

PR #127, opened 2024-04-21
Unify calling evaluation functions for individuals and populations

schmoelder:
Previously, there were two interfaces in the `OptimizationProblem` for calling evaluation functions (e.g. objectives): one for evaluating individuals, and one for populations.
These two methods were now unified.
To ensure backward compatibility, a 1D-Array is returned if a single individual is passed to the function.


schmoelder:
Since these are some bigger changes, I would like to keep them out of v0.9.1. However, 99cddeb closes #121, so we could decide to already include it in the new hotfix release.

ronald-jaepel (approved):

PR #130, opened 2024-04-22
Add qNParEGO Ax MOO Interface

ronald-jaepel:
As recommended by the `Ax` dev team, this PR adds the `qNParEGO` MOO interface that is useful for MOO problems with more than 7 objectives.

It currently fails the convergence tests labled `LinearConstraints`, `NonLinearConstraints` and `LinearNonLinearConstraints` because it has too many discrepancies between it's found solution and the optimal solution but passes all other tests.

### DIFF
--- a/CADETProcess/optimization/__init__.py
+++ b/CADETProcess/optimization/__init__.py
@@ -107,14 +107,14 @@ from .pymooAdapter import NSGA2, U_NSGA3
 import importlib
 
 try:
-   from .axAdapater import BotorchModular, GPEI, NEHVI
+   from .axAdapater import BotorchModular, GPEI, NEHVI, qNParEGO
    ax_imported = True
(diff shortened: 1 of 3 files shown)


ronald-jaepel:
@flo-schu Jo said you two had worked on qNParEGO but discarded it, do you remember anything about that?

flo-schu:
I have to admit, I don't remember that. The NEHVI algorithm is implemented, but if we did it, it must have been a very short interlude without significant progress. Sorry that I can't give you more than that.

ronald-jaepel:
Thanks for the quick reply! That's good news for me, it means there isn't some big memorable problem with qNParEGO. :)

ronald-jaepel:
Now all tests run. Ready for review and then merge.

PR #138, opened 2024-05-22
Fix documentation for AntiLangmuir isotherm

AntoniaBerger:
Fix incorrect declaration in antilangmuir as optional


schmoelder:
Hey @AntoniaBerger, since we discovered that the `antilangmuir` coefficient can technically take up values from [-inf, inf], I opted for a `SizedList(size='n_comp')`, i.e. removing the bounds imposed by a `SizedRangedList`. Please check again if everything works s.t. we can merge this.

PR #141, opened 2024-06-18
Adapt to new DG interface in CADET-Core 

schmoelder:
This PR updates CADET-Process to use the new DG interface.


CADETProcess/processModel/discretization.py
```diff
@@ -256,29 +251,11 @@ class LRMPDiscretizationDG(DGMixin):
...
         If False, use Jacobians generated by algorithmic differentiation (slower).
         Default is True.
-    reconstruction : Switch, optional
-        Method for spatial reconstruction. Valid values are 'WENO' (Weighted
-        Essentially Non-Oscillatory). Default is 'WENO'.
-    polynomial_degree : UnsignedInteger, optional
-        Degree of the polynomial used for spatial discretization. Default is 3.
     polydeg : UnsignedInteger, optional
-        Alias for polynomial_degree.
+        Degree of the polynomial used for spatial discretization. Default is 3.
```
jbreue16, line 255: default is 4 in cadet-core

CADETProcess/processModel/discretization.py
```diff
@@ -449,34 +418,16 @@ class GRMDiscretizationDG(DGMixin):
...
     polydeg : UnsignedInteger, optional
-        Alias for polynomial_degree.
-    polynomial_degree_particle : UnsignedInteger, optional
+        Degree of the polynomial used for axial discretization. Default is 3.
+    par_polydeg : UnsignedInteger, optional
         Degree of the polynomial used for particle radial discretization. Default is 3.
     exact_integration : Bool, optional
         Whether to use exact integration for the axial discretization.
         Default is False.
     exact_integration : Bool, optional
```
jbreue16, line 428: should be par_exact_integration
You could also leave this out and internally always set this to 1 since the other method is only feasible (wrt stiffness) in special cases

CADETProcess/processModel/discretization.py
schmoelder: thanks

CADETProcess/processModel/discretization.py
schmoelder: Good catch!
>  You could also leave this out and internally always set this to 1 since the other method is only feasible (wrt stiffness) in special cases
I think, we should mimic the interface of CADET-Core 1:1. Most users will not touch this anyway.

PR #142, opened 2024-06-18
Fix/use minutes

schmoelder:
Recently, the option to (not) plot the time axis using minutes was introduced.
This PR fixes some methods that were not properly implemented. Moreover, the name of the flag was changed from `use_minutes` to `x_axis_in_minutes` to make clear that only plotting is affected and not other parameter values (e.g. start and end times).

## Todo
- [x] Fix error (Error in type annotation)
- [x] Add flag to `plot_purity` (and check other plot methods again)

### DIFF
--- a/CADETProcess/comparison/comparator.py
+++ b/CADETProcess/comparison/comparator.py
@@ -271,24 +271,24 @@ class Comparator(Structure):
 
     def plot_comparison(
             self,
-            simulation_results: list[SimulationResults],
+            simulation_results: SimulationResults,
             axs: Axes | list[Axes] | None = None,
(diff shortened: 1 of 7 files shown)


AntoniaBerger (commented):
- One small thing: In CADETProcess/dynamicEvents/section.py line 437: 
We could also change the label of the x axis depending on the x_axis_in_minutes flag.

- One thought: In CADETProcess/dynamicEvents/section.py line 426-434:
If we get y from time in sec and then change time to minutes (430) do we loose the correct relations? Like in CADETProcess/dynamicEvents/event.py line799-819.

PR #154, opened 2024-07-03
Fix plot_at_position

hannahlanzrath:
This PR fixes plot_at_position on the dev branch by adding necessary parameters to the function definition.

### DIFF
--- a/CADETProcess/solution.py
+++ b/CADETProcess/solution.py
@@ -959,9 +959,12 @@ class SolutionBulk(SolutionBase):
     def plot_at_position(
             self,
             z: float,
-            components=None,
-            layout=None,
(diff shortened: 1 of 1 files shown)


hannahlanzrath:
@schmoelder If you give me a thumbs up I would let this automatically merge

CADETProcess/solution.py
```diff
@@ -971,10 +974,19 @@ def plot_at_position(
         ----------
         z : float
             Position for plotting.
+       start : float, optional
```
schmoelder, line 977: Check indentation here.

hannahlanzrath: Thank you! 🙏

PR #152, opened 2024-06-28
Optimization improvements

schmoelder:
This PR includes a couple of improvements to the optimization module:

## General improvements
- [x] Add option to disable progress report (which can take a lot of time)
- [x] Speed up cache pruning
- [x] Fix bug when checking linear equality constraints
- [x] Fix bug when adding linear and linear equality constraints
- [x] Improve testing performance by using in-memory cache
- [x] Update testing parameters
- [x] Add type hints (WIP)

Note, since the PR got too big, it was split and will be continued in #167

PR #158, opened 2024-07-11
Fit surrogate model on existing population

schmoelder:
This PR implements a `Surrogate` class for fitting GPs on existing data from optimizations.

Supersedes #45 

## To do
Note, there is some WIP in #152 which will also affect this PR. Should we already rebase onto that branch s.t. we can adapt the corresponding interfaces but risk some friction should there be some more changes upstream?

- [x] Fit GP on existing population
- [x] Provide method to evaluate objective functions, constraint functions etc.
- [ ] Validate surrogate model accuracy with (simple) test cases
- [ ] Test if optimizers converge to true solution using surrogate 

### Alternative surrogate modeling approches

Currently, only GPs are implemented. However, other surrogate models can be envisioned (e.g. ANNs). To improve a more modular architecture, we could subclasses a `SurrogateBase`.

## Follow-up projects

Once this is merged, we can also start working on other features which would improve or apply the surrogate models. Eventually, these should be moved to their own issues / PRs but for now, this is just a collection of ideas.

### The interface

Currently, the `SurrogateModel` class somewhat mimics an `OptimizationProblem` as it also provides methods for estimating objectives, nonlinear constraints etc. To demonstrate this, compare the `OptimizationProblem`
```mermaid
sequenceDiagram
    User->>+OptimizationProblem: evaluate_objectives(x)
    OptimizationProblem->>+User: f(x)
```

with the `SurrogateModel`
```mermaid
sequenceDiagram
    User->>+SurrogateModel: estimate_objectives(x)
    SurrogateModel->>+User: f*(x)
```

However, it is important to note that the `SurrogateModel` will never provide all functionality of the `OptimizationProblem`, such as specifying variables, constraints etc. Hence, it cannot directly be used as an `OptimizationProblem` e.g. to interface with an `Optimizer`.

To me, this means we should rethink the architecture and consider what exactly does the `SurrogateModel` replace? In the context of an `OptimizationProblem`, I would say, it actually replaces the evaluation toolchain (that which returns the values x->f/g/m/...). 

Consequently, we should consider moving the evaluation toolchain from the `OptimizationProblem` to its own module (which in the process would also make the `OptimizationProblem` less of a "god class" and would even allow reusing the toolchain in other places) and introduce an `EvaluationInterface`.

The architecture would then look something like the following:

```mermaid
sequenceDiagram
    User->>+OptimizationProblem: evaluate_objectives(x)
    OptimizationProblem->>+EvaluationInterface: evaluate(x)
    EvaluationInterface->>+OptimizationProblem: f(x)
    OptimizationProblem->>+User: f(x)
```

Where the `EvaluationInterface` is then implemented by both the `EvaluationPipeline` (i.e. the current "toolchain") and the `SurrogateModel`:

```mermaid
classDiagram
    class OptimizationProblem {
        evaluate_objectives(np.ndarray): np.ndarray
    }
    OptimizationProblem "1" *-- "1" EvaluationInterface
    
    class EvaluationInterface {
        <<interface>>
        +evaluate(np.ndarray): np.ndarray
    }
    
    SurrogateModel <|-- EvaluationInterface
    EvaluationPipeline <|-- EvaluationInterface
```

### Conditioned optimization problems

One of the original ideas for this projects came from optimization problems where we want to fix the value of one of the variables and then run the optimization to find the best point given this value. 

For this purpose, we should implement a `ConditionedOptimizationProblem` which wraps the original `OptimizationProblem` and provides an interface where the fixed variables are removed. While this is trivial to implement for bound constrained problems, it becomes potentially more complicated for problems with linear and nonlinear constraints.

### Plots

For process design, we are often not really interested in just the optimal point but in the general topology of the parameter space. E.g. we are interested in the contours of regions with a given purity. Finely sampling the parameter space would be very expensive so the idea could be to use a surrogate model for this purpose. See also [partial dependence plots](https://scikit-learn.org/stable/modules/partial_dependence.html) and #33.


schmoelder:
## Notes from Call with @maxsiska

### Normalization / StandardScaler
- Check if normalization allows for negative values
- Check if log-normalizaiton is possible (note, might become an issue for linear constraints)
- Check if inverse_transform (when estimating eval_functions) for `return_cov is True` works independent of used `Scaler`

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,360 @@
...
+        -------
+        tuple[GaussianProcessRegressor, StandardScaler, StandardScaler]
+            A tuple containing the gaussian process regressor, as well as scalers for
+            input and output dimensions.
+
+        """
+        X_scaler = StandardScaler().fit(X)
+        Y_scaler = StandardScaler().fit(Y)
+
+        gpr = GaussianProcessRegressor()
```
schmoelder, line 78: Consider allowing specification of hyper parameters via `kwargs` (with reasonable defaults).

See also: https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html

Important parameters:
- `kernel`: default is `ConstantKernel(1.0, constant_value_bounds="fixed") * RBF(1.0, **length_scale_bounds="fixed")` (consider moving to separate method, or expose their parameters in signature)
- `alpha`: Parameter to handle noisy data, default=1e-10

CADETProcess/optimization/surrogate.py
```diff
@@ -0,0 +1,360 @@
...
+        """Decorate function to ensure X array is an ndarray with ndmin=2."""
+        @wraps(func)
+        def wrapper(
+                self,
+                X: np.ndarray,
+                *args, **kwargs
+                ) -> Any:
+
+            X = np.array(X)
+            X_2d = np.array(X, ndmin=2)
```
schmoelder, line 181: Technically, this does not "ensure 2D" (only ndmin=2). Consequently, we should instead reshape to really ensure 2d.

Also, consider moving function out of class since it duplicates what is done in the `OptimizationProblem` class.

tests/test_surrogate.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_objectives(test_problem):
+    surrogate = create_surrogate(test_problem)
+
+    # TODO: Evaluate other points than the ones which were used for training.
+    X_test = surrogate.population.feasible.x[0:2]
+
+    F_test = surrogate.population.feasible.f[0:2]
+    F_est = surrogate.estimate_objectives(X_test)
```
schmoelder, line 69: Here, we need to use a different validation set to test if surrogate does predict well enough.

PR #167, opened 2024-08-06
Optimizer improvement v2

schmoelder:
This PR includes a couple of improvements to the optimization module and is a continuation of #152

## General improvements
- [x] Consider bound and linear constraint violations in feasibility
- [x] `Individual.cv` does not work if any of the bounds are `None` (breaks `np.concatenate`).
- [ ] Add tests for dependent variables

## Open questions
- [x] `cv_tol` was renamed to `cv_nonlincontol` which is more explicit, especially for reporting results in the `Individual` (since we also want to report the violations of other constraints). However, `cv_tol` is also a parameter for the `OptimizerBase`. Here, it's not always clear to me for what the `cv_tol` is used (I presume it's also mostly just for nonlinear constraints but I'll have to investigate this). Also, does this sacrifice a breaking change? Generally, I would encourage not using `x_tol`, `f_tol` or `cv_(nonlincon)tol` but rather setting the parameters as described in the corresponding optimizer implementation (e.g. `gtol` in `TrustConstr`).

### DIFF
--- a/CADETProcess/optimization/axAdapater.py
+++ b/CADETProcess/optimization/axAdapater.py
@@ -25,6 +25,7 @@ from botorch.acquisition.analytic import (
     LogExpectedImprovement
 )
 
+from CADETProcess import CADETProcessError
 from CADETProcess.dataStructure import UnsignedInteger, Typed, Float
 from CADETProcess.optimization.optimizationProblem import OptimizationProblem
 from CADETProcess.optimization import OptimizerBase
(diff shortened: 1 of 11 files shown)


schmoelder:
@ron, I think it's working now. I also tested the Püttmann [sic!] study and it doesn't throw any errors. Would you mind checking again? It's generally been reviewed by @flo-schu, so a sanity check should suffice.

ronald-jaepel (approved):

ronald-jaepel:
looks good to me! 👍

PR #170, opened 2024-09-26
Fix loading of multi-cycle solutions

ronald-jaepel:
This PR fixes a bug that occurs when loading the results of multi-cycle simulations where only the first cycle gets loaded for all cycles.

The cause was a block of code that was not indented enough. The for loop iterates over all cycles. The codeblock in question contained the code to bump the start-coordinate, so that the next cycle would be loaded on the next iteration of the for loop.

simplified example before:
```python
start = 0
for cycle in cycles:
    end = start + len(time)
    if 'solution_x' in unit_solution.keys():
        load_solution_x(start, end)
    if 'solution_y' in unit_solution.keys():
        load_solution_y(start, end)
if 'solution_z' in unit_solution.keys():
    load_solution_z(start, end)
start = end - 1
```

example as it is now:

```python
start = 0
for cycle in cycles:
    end = start + len(time)
    if 'solution_x' in unit_solution.keys():
        load_solution_x(start, end)
    if 'solution_y' in unit_solution.keys():
        load_solution_y(start, end)
    if 'solution_z' in unit_solution.keys():
        load_solution_z(start, end)
    start = end - 1
```

### DIFF
--- a/CADETProcess/simulator/cadetAdapter.py
+++ b/CADETProcess/simulator/cadetAdapter.py
@@ -693,17 +693,17 @@ class Cadet(SimulatorBase):
                                 )
                             )
 
-                    if 'solution_volume' in unit_solution.keys():
-                        sol_volume = unit_solution.solution_volume[start:end, :]
-                        solution[unit.name]['volume'].append(
-                            SolutionVolume(
-                                unit.name,
-                                unit.component_system,
-                                time,
-                                sol_volume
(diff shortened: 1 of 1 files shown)


hannahlanzrath:
Funnily enough, I was looking at this exact spot, but didn't realize the issue was with the indentation. Big thanks for catching that!

Issue #98, opened 2024-02-14
Implement early stopping and batch trials for Ax


schmoelder:
Running batch trials is currently disabled in favor of early stopping, as both strategies were conflicting on Ax' side. Manually performing early stopping once the best-point/best-pareto can be returned from Ax.
- Improve obtaining best-point/best-pareto during optimization
- Improve setting termination criteria. For this it would be useful, if Ax could return the best point. This is currently not implemented
- Adjust the number of cores to the number of batch trials, once batch mode is enabled again

## Starting Point

This is a classmethod of AxInterface

```py
from ax.modelbridge.modelbridge_utils import (
    observed_pareto_frontier as observed_pareto,
    predicted_pareto_frontier as predicted_pareto,
)

def _get_best_results(self, modelbridge, optimization_config):
    raise NotImplementedError("does not work so far.")
    n_obj = self.optimization_problem.n_objectives 
    if n_obj > 1:
        pareto_optimal_observations = observed_pareto(
            modelbridge, 
            optimization_config=optimization_config
        )
        pareto_optimal_predictions = predicted_pareto(
            modelbridge, 
            optimization_config=optimization_config
        )

        # here also work needs to be done.

    else:
        arm, obj = modelbridge.model_best_point()
        x = np.array(list(arm.parameters.values()))
        metrics = optimization_config.objective.metric_names
        assert len(metrics) == 1, "In single objective problems, only 1 Metric is allowed."
        f = obj[0][metrics[0]]

        return np.array([x], ndmin=2), np.array([f], ndmin=2)
```

then in the optimization loop:

```py
X_opt, F_opt = self._get_best_results(
    modelbridge=modelbridge, 
    optimization_config=optimization_config
)
```

However, currently Ax does not return the best point, the reason for this is unclear

---

Here is another removed old snipped from code. May be another starting point for providing the current optimum to `self._post_processing` in the training loop.

```python
self._data = ax.Data.from_multiple_data([self._data, trial.fetch_data()])
new_value = trial.fetch_data().df["mean"].min()

print(
    f"Iteration: Best in iteration {new_value:.3f}, "
    f"Best so far: {self._data.df['mean'].min():.3f}"
)
```

flo-schu:
## General Ax usage

[...]

PR #179, opened 2024-11-18
Resolves: ImportError cannot import name FixedNoiseGP from `botorch.models.gp_regression`

flo-schu:
FixedNoiseGP was deprecated for a while. The PR replaces `FixedNoiseGP` with `SingleTaskGP`. If ax tests pass, this should be ready to merge. SingleTaskGP has already been used in #98


schmoelder (changes requested):
Thanks @flo-schu for fixing this! 

Before merging, you could also try if this runs with the latest version of `ax` which is currently pinned to `<0.4.3`  (see https://github.com/fau-advanced-separations/CADET-Process/blob/dev/pyproject.toml#L59) to avoid crashes.

Note, the CI currently does not run the optimization tests. So we have to run those locally. I know, not ideal but that's just how it is for now. :sweat_smile:

PR #182, opened 2024-11-27
Add CITATION.bib

jbreue16:
The CITATION.bib file serves as an collection of papers that we like to have cited when CADET-Process is used. Further, it is recognized by github and added to the sidebar of the repo.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,27 @@
+{
+  "title": "CADET-Process",
+  "upload_type": "software",
+  "creators": [
+
+    {
+      "name": "Schmölder, Johannes",
+      "orcid": "0000-0003-0446-7209",
+      "affiliation": "Forschungszentrum Jülich"
+    },
+    {
+      "name": "Jäpel, Ronald",
+      "orcid": "0000-0002-4912-5176",
+      "affiliation": "Forschungszentrum Jülich"
+    }
+  ],
+  "license": "GPL-3.0",
+  "keywords": [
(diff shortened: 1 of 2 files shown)


jbreue16:
Ideally, this file would be a centralized bib, i.e. reused in the readme and the documentation

.zenodo.json
```diff
@@ -0,0 +1,27 @@
+{
+  "title": "CADET-Process: Version 0.9.2",
```
schmoelder, line 2: @hannahlanzrath is this already the file for the next version?

schmoelder:
> Ideally, this file would be a centralized bib, i.e. reused in the readme and the documentation

So you mean, this would then replace the one we have in the documentation? But there are also many references to publications that aren't ours... But maybe I don't fully get where you're going with this. :sweat_smile:

jbreue16:
> So you mean, this would then replace the one we have in the documentation? But there are also many references to publications that aren't ours... But maybe I don't fully get where you're going with this. 😅

Yes that's what I meant. But you're right, this can only a bib for cadet-process publications, maybe there'll be more than one in the future. So there's probably a centralized bib file somewhere else already and we don't need to further change anything.

.zenodo.json
schmoelder: ok, next version will be 0.10.0

hannahlanzrath: I will change, amend and force push if you're ok with that

PR #191, opened 2024-12-03
fixed comp sys bug with molecular weights, added test

daklauss:
Fixed the bug and added unittest to check molecular weight.

Renamed to singular: molecular_weights -> molecular_weight because species used  molecular_weight and in all __init__ methods, keywords are simply passed to the next instance -> all instance ( component, species, component_system) name their attributes with the same names.

Maybe only use plural, otherwise need to convert somwhere in instantiation?

attrs could use aliases ;)


CADETProcess/processModel/componentSystem.py
```diff
@@ -182,7 +182,7 @@ class ComponentSystem(Structure):
     name = String()
 
     def __init__(
-            self, components=None, name=None, charges=None, molecular_weights=None):
+            self, components=None, name=None, charges=None, molecular_weight=None):
```
schmoelder, line 185: Let's be consistent and use plural (`molecular_weight*s*`) for the `ComponentSystem` and singular for the Component / `Species`

schmoelder (changes requested):
Thanks for fixing this. Please incorporate my review and improve your git commit message. For some guidance, refer to [this article](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages).

Issue #144, opened 2024-06-19
Monitor test coverage


schmoelder:

schmoelder:
When measuring coverage, I think it's fine to only check the `CADETProcess` directory. No need to check examples, docs or the tests themselves.

PR #192, opened 2024-12-03
Change unittest to pytest

daklauss:
- [x] Changed unittest to pytest
- [x] Use ```pytest.mark.slow``` to ignore to complicated tests

removed - [ ] Using [Coverage](https://coverage.readthedocs.io/en/7.6.8/) for coverage report to also cover #144

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -68,7 +68,7 @@ jobs:
 
       - name: Test
         run: |
-          python -m unittest discover -s tests
+          pytest tests -m "not slow" --durations=0
 
       - name: Install pypa/build
         run: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,6 @@ docs = [
     "sphinx_copybutton>=0.5.1",
     "sphinx-sitemap>=2.5.0",
 ]
-
 ax = [
     "ax-platform >=0.3.5"
 ]
@@ -77,4 +76,10 @@ line-length = 88
 [tool.pytest.ini_options]
 testpaths = [
     "tests",
+]
+pythonpath = [
+    "tests",
+]
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
 ]
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -11,7 +11,7 @@
 Notes
 -----
 Since the EventHandler defines an interface, that requires the
-implementation of some methods, a TestHandler class is defined.
+implementation of some methods, a HandlerFixture class is defined.
 
 Maybe this is too complicated, just use Process instead?
 """
@@ -31,7 +31,8 @@ from CADETProcess.dataStructure import (
 plot = True
 
 
-class TestPerformer(Structure):
+class PerformerFixture(Structure):
+
     n_entries = 4
     n_coeff = 4
 
@@ -79,10 +80,11 @@ class TestPerformer(Structure):
         return parameters
 
 
-class TestHandler(CADETProcess.dynamicEvents.EventHandler):
+class HandlerFixture(CADETProcess.dynamicEvents.EventHandler):
+
     def __init__(self):
         self.name = None
-        self.performer = TestPerformer()
+        self.performer = PerformerFixture()
         super().__init__()
 
     @property
@@ -100,7 +102,7 @@ class TestHandler(CADETProcess.dynamicEvents.EventHandler):
         except KeyError:
             pass
 
-        super(TestHandler, self.__class__).parameters.fset(self, parameters)
+        super(HandlerFixture, self.__class__).parameters.fset(self, parameters)
 
     @property
     def section_dependent_parameters(self):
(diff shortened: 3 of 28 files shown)


schmoelder:
For reference, this is how we do it in CADET-Python: https://github.com/cadet/CADET-Python/blob/master/.github/workflows/pipeline.yml

daklauss:
Removed coverage for a later issue #144. 
grafik
 https://github.com/fau-advanced-separations/CADET-Process/actions/runs/12177180699/job/33964452874

Pytest with coverage as i used needs way to long for a simple push (double time). Considering using it just on merge or action. 

Pytest alone should work.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
 testing = [
     "certifi", # tries to prevent certificate problems on windows
     "pytest",
+    "coverage",
```
schmoelder, line 45: please move to coverage PR

schmoelder:
This one here could also be marked as `slow` (for now):
https://github.com/fau-advanced-separations/CADET-Process/blob/ce627e2e6b94b0a26840ddc119abea645933e776/tests/test_cadet_adapter.py#L157

tests/test_events.py
```diff
@@ -117,9 +120,6 @@ def polynomial_parameters(self):
 
 class Test_Events(unittest.TestCase):
 
-    def __init__(self, methodName='runTest'):
```
schmoelder, line 120: short question, is `methodName='runTest'` also the default for `unittest.TestCase`? Otherwise, we would have to check whether this changes any of the expected testing behavior.

daklauss: As far as i undestand the [ docs ](https://docs.python.org/3/library/unittest.html#unittest.TestCase), ´´´´methodName='runTest'´´´ is the default. But we use this constructor with pretty much all test cases. So i'm not sure how many tests are ignored.

daklauss: With `methodName='runTest'` as default, we should remove this constructor either way because it is duplicated.

schmoelder: good. if it's the default anyway, then we can safely remove it.

daklauss: Done

daklauss:
Still many warnings, but more or less the same as in dev. Tests run overall slightly slower compared to unittest. Otherwise as in [dev](https://github.com/fau-advanced-separations/CADET-Process/actions/runs/12354318876/job/34475399971), now 220 tests were run, 3 skipped.

schmoelder:
Any new warnings or simply the same as with `unittest`?

.github/workflows/pipeline.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
       - name: Test
         run: |
-          python -m unittest discover -s tests
+          pytest tests -m "not slow" --durations=0
```
schmoelder, line 71: Just out of curiosity, what does the `durations=0` flag do?

daklauss:
> Any new warnings or simply the same as with `unittest`?
As far as i can tell simply the same as with `unittest`. Pytest counts all warnings thrown in a run what unittest does not, so i don't have a metric there, but i have found the same warnings in the dev runs. Still found some `__init__`....

pyproject.toml
```diff
@@ -78,3 +77,9 @@ line-length = 88
 testpaths = [
     "tests",
 ]
+pythonpath = [
+    "tests",
```
schmoelder, line 81: I think, an alternative would be to run 
```
pytest tests --rootdir=tests
```
then, you probably don't need to add this here.

But in the end, I don't really think it's important. 

Do you have a preference?

tests/test_events.py
```diff
@@ -32,6 +32,7 @@
 
 
 class TestPerformer(Structure):
+    __test__ = False
```
schmoelder, line 35: Could you please explain what this does?

schmoelder (commented):
LGTM, just some minor comments / Changes.

.github/workflows/pipeline.yml
daklauss: It prints the times of every tests into the log. I was a bit curious about how long each test needs. `durations=n` would just print the n slowest tests.

tests/test_events.py
daklauss: Pytest tries to discover this as a TestClass because the Name starts with `Test`. But `TestPerformer` is a structure class used by the following tests. With `__test__ = False` pytest doesn't consider this class to be a Test and does not throw a warning because of overwritten constructor.

Maybe renaming would be the better option here.

pyproject.toml
daklauss: I prefer it as a pytest option in the toml. I guess it's easier to configure pytest in your ide when this option is set, but could be wrong here.

.github/workflows/pipeline.yml
daklauss: grafik

pyproject.toml
schmoelder: Good point.

schmoelder:
Alrighty, looks good to me. Please rebase and clean up the git history, then we can merge.

tests/test_events.py
daklauss: Maybe next time

daklauss: I would rename them for convention issues. Naming them `TestPerformer` and `TestHandler` and then telling pytest they arent tests is a bit weird. But i'm not sure how to rename them, also because i'm not sure what they are doing exactly.

daklauss: Thanks to the great schmoelder, i was able to rename them into PerformerFixture and HandlerFixture.

daklauss:
> Alrighty, looks good to me. Please rebase and clean up the git history, then we can merge.

Done

PR #206, opened 2025-01-08
Add coverage pipeline

### LINKED ISSUE #144: Monitor test coverage

daklauss:
Pull Request to add coverage pipeline to have a tool to measure test coverage. Using [Coverage](https://coverage.readthedocs.io/en/7.6.10/)
Closes #144 

Imo CI/CD should only activate manualy or when pushing to dev or master, because coverage tests leading to a significant time increase (like doubled) for tests.

Maybe an option to disable slow tests, otherwise every test should be run.

## Next steps
- [x] Check if it's possible to only run coverage for changed modules
- [x] Add flags to run pipeline with specific settings / on specific branches (ask @jbreue16)
- [x] Compare pipeline to other modules (e.g. networkx, pymoo, ...)
- [x] Ask for feedback by Michael Osthege
- [x] Check why some tests are slow, try to reduce runtime (e.g. reduce number of generations)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,76 @@
+name: coverage
+
+on:
+  push:
+    branches:
+      - master
+      - dev
+      - 144-monitor-test-coverage
+  workflow_dispatch:
+      inputs:
+        slow:
+          type: boolean
+          description: Run with slow tests
+          default: false
+
+jobs:
+  coverage-job:
+    runs-on: ${{ matrix.os }}
+
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.12"]
+
+    env:
+      CONDA_FILE: environment.yml
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Get Date
+        id: get-date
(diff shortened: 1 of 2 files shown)


schmoelder:
@daklauss you requested a review but there are still many open points. Should these become separate follow-ups or do you want to address them in this PR?

daklauss:
> @daklauss you requested a review but there are still many open points. Should these become separate follow-ups or do you want to address them in this PR?
@schmoelder i requested this review last week because i wanted an opinion that where adressed in the dev-call. I still want to adress the open points, but i think they won't be done in this PR. I forked CADET-Process so i am able to actually test ```workflow_dispatch``` runs. So i guess this PR will be sooner or later closed without merging. Will do this most likely next week.

daklauss:
In comparison, coverage on 3.12 needs about 10minutes, test-job alone only 6m. So coverage needs a lot more time. Otherwise test-pipeline for macos needs abouter 11m.  Maybe we could consider using test-job only on branches and coverage only on dev/main?

Concerning failing tests; coverage only applies on push on dev/main, so usually after merging a pull request. I don't think it matters that much to ignore the failing coverage pipeline because we only merge after the tests already passed in PRs. Otherwise we wouldn't get to know if e.g. pytest interface changed or something like that.

schmoelder:
Sounds reasonable.

PR #211, opened 2025-01-14
Make run methods private

### LINKED ISSUE #194: Make Simulator.run() function private
To simulate processes, the user should use `simulate()`, which also performs configuration checks before calling the `run()` method by the `SimulatorImplementation` (e.g. `CadetAdapter`).

To fix this, make the `run()` method private (i.e. rename to `_run()`), which then needs to be implemented by `CadetAdatper`.

Note, the same thing should be done for the `OptimizerBase.run` method.

schmoelder:
Fixes #194


hannahlanzrath (commented):
After the amend, all tests are running. 👍

PR #213, opened 2025-01-22
Add warning to simulator when setting n_cycles_max below n_cycles_min

ronald-jaepel:
A user on the forum was confused why n_cycles_max was not working as expected so I added this warning to clarify the behavior to future users.

### DIFF
--- a/CADETProcess/simulator/simulator.py
+++ b/CADETProcess/simulator/simulator.py
@@ -51,7 +51,8 @@ class SimulatorBase(Structure):
 
     n_cycles = UnsignedInteger(default=1)
     evaluate_stationarity = Bool(default=False)
-    n_cycles_min = UnsignedInteger(default=5)
+    n_cycles_min = UnsignedInteger(default=1)
+    n_cycles_batch = UnsignedInteger(default=5)
     n_cycles_max = UnsignedInteger(default=100)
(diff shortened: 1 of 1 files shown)


CADETProcess/simulator/simulator.py
```diff
@@ -248,6 +248,13 @@ def simulate(self, process, previous_results=None, **kwargs):
         if not process.check_config():
             raise CADETProcessError("Process is not configured correctly.")
 
+        if self.n_cycles_max < self.n_cycles_min:
```
schmoelder, line 251: We could consider this an invalid configuration and throw an error.

schmoelder:
We could also reconsider if we want to continue using `n_cycles_min` also for the number of cycles simulated per batch. I did this originally because processing the results of a cycle creates a lot of overhead so it was better to simply run multiple cycles at once, even though it was maybe sometimes more than required. This might have changed with the new C-API.
Or, we could simply add a new setting.

@daklauss Do you want to profile this?

ronald-jaepel:
I've changed the warning to an error and added a new parameter `n_cycles_batch` for the batching of cycle simulation calls. Now `n_cycles_min` must be lower than `n_cycles_max` and `n_cycles_batch` is set to `n_cycles_max` if it is larger silently.

PR #216, opened 2025-01-27
Fix Aggregator setter

### LINKED ISSUE #215: Setting of stochiometric matrix in optimization Problem confusing or broken
From the Forum, there's a use-case where the user wants to estimate the stochiometric matrix of a reaction using CADET-Process. 

The problem could be a bug or a misconfiguration.

Here's an MRE that crashes:

```Python
from CADETProcess.processModel import ComponentSystem, MassActionLaw, Cstr, FlowSheet, Process
from CADETProcess.simulator import Cadet
from CADETProcess.optimization import OptimizationProblem, U_NSGA3

component_system = ComponentSystem(["glucose", "glycerol", "NH3", "CDW"])

reaction_system = MassActionLaw(component_system)
reaction_system.add_reaction(
    indices=[0, 1, 2, 3],
    coefficients=[-1.0, -1.0, -1.0, 1.0],
    k_fwd=1.0,
    k_bwd=0.0
)

reactor = Cstr(component_system, "reactor")
reactor.bulk_reaction_model = reaction_system
reactor.V = 1.0
reactor.c = [1.0, 1.0, 1.0, 0.0]

flow_sheet = FlowSheet(component_system, "flow_sheet")
flow_sheet.add_unit(reactor)

process = Process(flow_sheet, "reaction_fermentation")
process.cycle_time = 180

optimization_problem = OptimizationProblem("fermentation_optimization")
optimization_problem.add_evaluation_object(process)

optimization_problem.add_variable(
    name="stoichiometric_coefficient",
    parameter_path="flow_sheet.reactor.bulk_reaction_model.stoich",
    indices=[0],
    lb=-5.0,
    ub=5.0
)

simulator = Cadet()
optimization_problem.add_evaluator(simulator)


def callback(simulation_results, individual, evaluation_object, callbacks_dir='./'):
    simulation_results.solution.reactor.outlet.plot(
        file_name=f'{callbacks_dir}/{individual.id}_{evaluation_object}_comparison.png',
        show=False
    )


optimization_problem.add_callback(callback, requires=[simulator])

optimization_problem.evaluate_callbacks([2])


```

With the error: 
```
There was an exception in simulate_n_cycles
Traceback (most recent call last):
  File "C:\Users\ronal\PycharmProjects\CADET-Process\CADETProcess\log.py", line 172, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ronal\PycharmProjects\CADET-Process\CADETProcess\simulator\simulator.py", line 315, in simulate_n_cycles
    return self._run(process, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ronal\PycharmProjects\CADET-Process\CADETProcess\simulator\cadetAdapter.py", line 133, in wrapper
    results = func(self, process, *args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ronal\PycharmProjects\CADET-Process\CADETProcess\simulator\cadetAdapter.py", line 246, in _run
    raise CADETProcessError(
CADETProcess.CADETProcessError.CADETProcessError: CADET Error: Simulation failed with ERROR: Size of field MAL_STOICHIOMETRY_BULK must be a positive multiple of NCOMP (4)
```

The problem stem from (I think) `dataStructure.py` line 155:
https://github.com/fau-advanced-separations/CADET-Process/blob/a4e2d46719e108202e4b8395d29dcdcf70aec6a7/CADETProcess/dataStructure/dataStructure.py#L153-L156

where we get the `value` from the setter function as `value = [2, -1, -1, 1]` but then set an indexed subset of that array into the `self.parameter_name`.

schmoelder:
Fixes #215 

Added methods to indicate whether the internal data should be transposed (i.e., have shape `(n_containers, *parameter_shape))`. While this resolves the immediate issue, I had to overwrite the `__set__` method of the Aggregator base class, which feels less than ideal. I'll revisit this implementation to explore alternative approaches, but it should work for now.

Additionally, I noticed there are currently no tests for any aggregated parameters. This is unexpected, as I was fairly certain I had written some. Maybe they got lost somewhere...

ToDos:
- [x] Stoich coeffs in reactions should be sized
- [x] Check if the wrong code snippet is commented out in `SizedAggregator.__set__`
- [x] Check length of container vs value in `Aggregator.__set__`
- [x] Check consistency of transpose=True vs False in `SizedAggregator`
- [x] Add tests

### DIFF
--- a/CADETProcess/dataStructure/aggregator.py
+++ b/CADETProcess/dataStructure/aggregator.py
@@ -6,9 +6,29 @@ from .dataStructure import Aggregator
 class SizedAggregator(Aggregator):
     """Aggregator for sized parameters."""
 
+    def __init__(self, *args, transpose=False, **kwargs):
+        """
+        Initialize a SizedAggregator instance.
+
+        Parameters
+        ----------
+        *args : Any
+            Variable length argument list.
+        transpose : bool, options
+            If False, the parameter shape will be ((n_instances, ) + parameter_shape).
+            Else, it will be (parameter_shape + (n_instances, ))
+            The default is False.
+        **kwargs : Any
+            Arbitrary keyword arguments.
+        """
+        self.transpose = transpose
+
+        super().__init__(*args, **kwargs)
+
     def _parameter_shape(self, instance):
-        values = self._get_parameter_values_from_container(instance)
-        shapes = [np.array(el, ndmin=1).shape for el in values]
+        values = self._get_parameter_values_from_container(instance, transpose=False)
+
+        shapes = [el.shape for el in values]
 
         if len(set(shapes)) > 1:
             raise ValueError("Inconsistent parameter shapes.")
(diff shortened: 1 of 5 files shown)


schmoelder:
I refactored the transport functionality and now only the `__get__` and `__set__` methods are extended in the subclass (instead of being overwritten) which I find much more elegant.

I also added tests for the aggregator, so once tests pass, I'd appreciate another thorough review. :nerd_face:

ronald-jaepel:
Usage in optimizationProblems works nicely. 👍

Setting slices of aggregators didn't work. It didn't raise an Error or Warning and also didn't change the values. I've extended the tests for that. If that is a use-case you don't see users using, feel free to remove the tests again.

schmoelder:
> Setting slices of aggregators didn't work. It didn't raise an Error or Warning and also didn't change the values. I've extended the tests for that. If that is a use-case you don't see users using, feel free to remove the tests again.

Damn, that's a tricky one, since we're always returning a new value that is being constructed by iterating over the internal instances. So there is no internal reference that points back to the underlying container elements. I'm honestly not completely sure how to handle this but I'll think a bit further about it.

schmoelder:
Puh, this was a "difficult birth". In the end I did decide to provide proxy objects to the aggregated parameters which now seems to work properly. I had to adapt one more test but I think it's fine now...

ronald-jaepel (approved):

PR #218, opened 2025-02-07
Improve interpolation of concentration profiles

schmoelder:
This PR adds the option to specify the desired interpolator when automatically adding Events for concentration or flow rates to a `Process`. Moreover, it ensures a sufficient number of time points to be sampled from the interpolated signal. 

## To do:
- [ ] Unify flow profile and concentration profile?
- [x] Add tests

### DIFF
--- a/CADETProcess/processModel/process.py
+++ b/CADETProcess/processModel/process.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
-from warnings import warn
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


schmoelder:
@hannahlanzrath if I remember correctly, we had some discussions about this feature previously, right? Would you mind reviewing and if possible had some tests?

Cheers

hannahlanzrath (approved):

Issue #19, opened 2023-03-11
Use Component names when adding Fractions


schmoelder:
Currently, component indices need to be specified when adding fractions in the Fractionator.

Instead, it would be more coherent to use component names.

PR #219, opened 2025-02-07
[WIP] Overhaul Fractionator

### LINKED ISSUE #56: Improve FractionationOptimization
During collaboration with @T-Fuchs, we found some issues/ideas that should be addressed to improve the automatic fractionation.

## `c_min`
A minimum concentration `c_min` is used for calculating local purity profiles which in turn are used for determining the initial values of the optimization problem.
The attribute has a default value of $1\cdot 10^{-3}~mM$ and lives within the `SolutionIO` class. While this is works for most cases, it would be useful to allow users to specify a custom `c_min`, specifically for fractionation. Especially for large molecules such as proteins, it might make sense to chose different values. This should be handled by the `FractionationOptimization` class.

## Fractionation Window
Mostly, it's hard to predict, when the elution starts. However, sometimes there are some pre-peaks (e.g. from overloading the column) that we always want to avoid. Since the `Fractionator` would also try to handle these peaks, this makes the optimization problem unnecessarily complicated. By adding a fractionation window (e.g. `start`, and `end`, similar to the `Comparator`), this could be avoided.

## Normalization
Normalization is currently not supported for fractionation, but can be advantageous for concentration profiles with values >> 1 or << 1.

Currently, when calling `SolutionIO.normalize()`, each component profile is normalized to $1$ using its own maximum value. While this is useful for situations such as shape comparison with reference data, it's problematic when determining the mass/purity of components.
There should be options to:
- Normalize all of the profiles to the max of all components (proportional scaling), or
- Continue normalizing each profile to 1 but keeping track of normalization factors to correct mass

## Optimizer Options and Failure Handling
I tried to find reasonable values for optimizer parameters (e.g. `Cobyla.rhobeg`) that work for most situations. However, it's almost impossible to find parameters that fit all situations.
It is already possible to modify these options in place; this should be made clearer in the docs.
Moreover, when an optimizer fails (e.g. fails to converge), this should be properly handled, e.g. raise a `CADETProcessError` that could be handled by either taking the best point so far, the initial values or, if `ignore_failed=False` simply raise an exception. Currently, this failure is silent which is not good practice.

schmoelder:
This PR has some general improvements for the fractionation module. However, it's still very much work in progress. We should also reconsider some aspects once we've decided how to proceed with the integration of Fields / StateStructure that was developed for CADET-Python-Simulator.

Potentially fixes #56, #19

### DIFF
--- a/CADETProcess/fractionation/fractionationOptimizer.py
+++ b/CADETProcess/fractionation/fractionationOptimizer.py
@@ -125,7 +125,7 @@ class FractionationOptimizer:
         Fractionator
             The Fractionator object that has been set up using the provided arguments.
         """
-        frac = Fractionator(
+        frac = Fractionator.from_simulation_results(
             simulation_results,
             components=components,
             use_total_concentration_components=use_total_concentration_components,
--- a/CADETProcess/fractionation/fractionator.py
+++ b/CADETProcess/fractionation/fractionator.py
@@ -1,4 +1,3 @@
-import os
 from collections import defaultdict
 from functools import wraps
 from typing import Any, Callable, Optional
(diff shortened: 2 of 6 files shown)


schmoelder:
Closing this in favour of a clean reimplementation from the current state of dev. The ProcessMeta work has already landed. The Fractionator API direction (taking chromatograms + ProcessMeta directly instead of SimulationResults) is still worth doing, but I want to start from the current codebase rather than resolve 16 months of conflicts. The new implementation will add a deprecation path for the SimulationResults-based constructor and extend optimize_fractionation to accept SolutionIO or list[SolutionIO] directly alongside SimulationResults.

schmoelder:
Superseded by #387

PR #221, opened 2025-02-14
Remove Height from Shape metric

schmoelder:
After showing deprecation warnings for some time, This PR finally removes the Height from the Shape metric. Users can still manually specify Height as a separate metric.


ronald-jaepel (approved):
LGTM

PR #223, opened 2025-02-14
Add/component names when adding reactions

### LINKED ISSUE #18: Use component names when adding reactions
Currently, component indices need to be specified when adding reactions to a reaction system.

Instead, it would be more coherent to use component names.

AntoniaBerger:
resolves #18


schmoelder:
There was a naming conflict of the `Callable` type from the `typing` module with our own `Callable` type of the `parameters` module. This was fixed by explicitly specifying `__all__` in the `deprecation` module (Something we should do more often, btw).

Before merging, we should also check the documentation (`docs/user_guide/process_model/reaction`).

PR #226, opened 2025-02-15
Chore

schmoelder:
This PR aggregates some chore, including
- Improved docstrings / type annotations
- Additional dunder methods
- A new `ChromatographicColumnBase` class
- Removed `spatial_methods` from discretization parameters (maybe we shouldn't even consider it a `parameter`)

### DIFF
--- a/CADETProcess/dataStructure/nested_dict.py
+++ b/CADETProcess/dataStructure/nested_dict.py
@@ -5,6 +5,7 @@ from operator import getitem
 __all__ = [
     "check_nested",
     "generate_nested_dict",
+    "update_dict_recursively",
     "get_nested_value",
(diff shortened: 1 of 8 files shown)


ronald-jaepel:
LGTM. 

What's the plan for the `ChromatographicColumnBase` class?

ronald-jaepel (approved):

schmoelder:
> What's the plan for the `ChromatographicColumnBase` class?

There are some situations, where I want to make sure that a unit operation is actually a chromatographic column (e.g. in the `CarouselBuilder` or in a newer tool for automatically estimating parameters. It should work with any of the chromatographic column models, but obviously not with a `TubularReactor`.

PR #242, opened 2025-03-13
Add ShapeFront difference metric

schmoelder:
This PR adds another difference metric originally implemented in CADET-Match. It is similar (no pun intended) to the Shape metric. However, it only accounts for the front of the peak which is useful for components which have a tendency for tailing, such as Blue Dextran.

### DIFF
--- a/CADETProcess/comparison/difference.py
+++ b/CADETProcess/comparison/difference.py
@@ -218,8 +218,8 @@ class DifferenceBase(MetricBase):
 
         return wrapper
 
-    def resamples_smoothes_and_normalizes_solution(func):
-        """Decorator to automatically smooth and normalize solution."""
+    def resamples_normalizes_and_smoothes_solution(func):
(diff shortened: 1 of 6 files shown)


CADETProcess/comparison/difference.py
```diff
@@ -530,7 +532,7 @@ def __init__(
             )
             self.reference_der_sliced = slice_solution(
                 self.reference_der,
-                None,
+                self.components,
```
schmoelder, line 535: To be honest, I'm not sure this is a good idea. This way, we "force" our Reference to have the same component system as our solution. While more consistent, this also comes with some downsides:
- it is a bit more cumbersome for the user to set up
- it could be a limitation when we are comparing to some sum-signal (e.g. total concentration)

so maybe we should reverse this change.

There were some issues during testing but maybe they came from somewhere else; I'll have to re-check...

PR #244, opened 2025-03-14
Add/use component names when adding fractions

### LINKED ISSUE #19: Use Component names when adding Fractions

AntoniaBerger:
resolves #19


schmoelder:
Good work @AntoniaBerger. I updated some of the type annotations and docstrings. Let's have another look together tomorrow so I can explain some of my reasoning.

PR #245, opened 2025-03-14
Adapt to CADET-Python's new run interface

ronald-jaepel:
This prepares for the new interface introduced in https://github.com/cadet/CADET-Python/pull/51

### DIFF
--- a/CADETProcess/simulator/cadetAdapter.py
+++ b/CADETProcess/simulator/cadetAdapter.py
@@ -337,7 +337,7 @@ class Cadet(SimulatorBase):
         if hasattr(cadet, "run_simulation"):
             return_information = cadet.run_simulation(timeout=self.timeout)
         else:
-            return_information = cadet.run_load()
+            return_information = cadet.run_load(timeout=self.timeout)
 
         return cadet
 
--- a/CADETProcess/simulator/cadetAdapter.py
+++ b/CADETProcess/simulator/cadetAdapter.py
@@ -173,7 +173,7 @@ class Cadet(SimulatorBase):
         else:
             return_information = cadet.run_load()
 
-        os.remove(lwe_hdf5_path)
+        cadet.delete_file()
 
         if return_information.return_code != 0:
             raise CADETProcessError(return_information)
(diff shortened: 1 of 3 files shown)


CADETProcess/simulator/cadetAdapter.py
```diff
@@ -232,7 +239,11 @@ def _run(self, process, cadet=None, file_path=None):
 
         try:
             start = time.time()
-            return_information = cadet.run_load(timeout=self.timeout)
+
+            if hasattr(cadet, "run_simulation"):
```
schmoelder, line 243: Here it would be useful to add an inline-comment that explains why this is done so that in the future, we can remove this.

CADETProcess/simulator/cadetAdapter.py
```diff
@@ -164,13 +164,20 @@ def check_cadet(self):
...
+        cadet.create_lwe(lwe_hdf5_path)
 
-        cadet_model.create_lwe(lwe_hdf5_path)
+        if hasattr(cadet, "run_simulation"):
+            return_information = cadet.run_simulation(timeout=self.timeout)
+        else:
+            return_information = cadet.run_load()
 
-        cadet_model.run()
         os.remove(lwe_hdf5_path)
```
schmoelder, line 176: here, we could also call the `delete_file` method of the cadet object.

schmoelder: Or, maybe even better, if we don't even add an `lwe_hdf5_path`, this is automatically deleted by the `create_lwe` method. This way, we would also future-proof us for `create_lwe` using the C-API.

schmoelder (changes requested):
Thanks for taking care of this. I like that you made it backwards compatible. As mentioned in one of the comments, it would be useful to add short comments s.t. we know this can be removed eventually (e.g. when we pin CADET-Python to a sufficiently high version).

CADETProcess/simulator/cadetAdapter.py
ronald-jaepel: If we don't supply `file_path=lwe_hdf5_path` to `create_lwe()`, it tries to create the `LWE.h5` file in the current working directory which leads to `IO Errors` both on my local tests and in the CI, probably due to access rights. I'll therefore still supply the `lwe_hdf5_path`.

CADETProcess/simulator/cadetAdapter.py
ronald-jaepel: Comment added. o7

PR #249, opened 2025-03-20
Improve GRM docstring clarity

ronald-jaepel:
This PR fixes two typos and improves the docstring clarity. Previously the GRM docstring referred to "reactor" concentrations. It now refers to just mobile and stationary phase concentrations.


schmoelder:
Good catch, thanks!

Issue #227, opened 2025-02-17
Feature request: Loading of previously generated optimization results


ronald-jaepel:
Covered in #228

PR #251, opened 2025-03-21
fixup! Adapt to CADET-Python's new run interface

schmoelder:
This PR fixes a small issue with CADET-Python's new load interface for H5 files.

Moreover, the `load_results` method is moved to the `OptimizationResults` (see #227).


schmoelder:
Puh, that was zäh...

PR #254, opened 2025-03-26
Use last point fulfilling purity constraints for initial values

schmoelder:
This PR updates the initial values routine of the Fractionator. Now, it starts the fraction at the first time point where the local purity exceeds the required purity and begins sending to waste starting from the last point where the purity is still above the required level (instead of the first point where it drops below). In practice, this change has minimal impact because we typically have enough data points to make the difference negligible. These are only initial values for the fractionation times, which are later optimized to ensure a cumulative required purity. However, the previous approach was slightly inconsistent, which was noted during testing.

Moreover, the default parameters for the COBYLA optimizer were updated since the initial step was much smaller than final tolerance. This now seems to produce more consistent results, as well as a better performance.

### DIFF
--- a/CADETProcess/fractionation/fractionationOptimizer.py
+++ b/CADETProcess/fractionation/fractionationOptimizer.py
@@ -55,9 +55,9 @@ class FractionationOptimizer():
         """
         if optimizer is None:
             optimizer = COBYLA()
-            optimizer.tol = 1e-3
+            optimizer.tol = 1e-4
             optimizer.catol = 5e-3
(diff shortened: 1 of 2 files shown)


daklauss (approved):
Comparing the old test [times](https://github.com/fau-advanced-separations/CADET-Process/actions/runs/14085938398/job/39449888125)
grafik
with the new test [times](https://github.com/fau-advanced-separations/CADET-Process/actions/runs/14084908724/job/39446299541)
grafik
seems to justify these changes and i hope this can be advanced further.
👍

PR #259, opened 2025-04-16
Add ruff linter+formater for unified coding style

daklauss:
Currently CADET-Process doesn't enforce unified coding style rules within the entire code. This is a first attempt  to define some coding style rules that we want to check with ruff.linter and maybe later even use ruff.format to automaticly enforce these rules.
First off some tasks have to be done:

- [x]  agree on linter rules
- [x]  only src of "CADET-Process" or also use the "tests"
- [x]  ruff/linter pipeline

### General Linting Rules

- **E and W**: These are typical Pycodestyle rules, addressing issues like the use of tabs mixed with spaces, and ensuring proper whitespace between operators, among others.
- **F**: These are Pyflakes rules, which check for issues related to imports, f-strings, unused imports, and more.
- **D**: This set of rules pertains to docstrings. Docstrings are enforced for all public classes, functions, methods, and packages, with the exception of modules. Rules `D100`, `D212`, and `D203` are excluded. We enforce a descriptive imperative sentence at the start of every docstring. Module docstrings are excluded due to compatibility issues with Sphinx docstrings that do not conform to standard rules. To exclude a docstring from Ruff checking, use `# noqa`.
- **ANN**: This rule enforces type checking for nearly all public and private functions and methods. Rule `ANN401` is excluded because there are many cases where using `Type[Any]` is more convenient. We may include this rule at a later time.
- **I**: This rule ensure that imports are correctly sorted and grouped according to the specified categories: standard library, third-party, and first-party imports."
- For all test files, docstring and annotation rules are disabled, as well as `E731` and `E402` (module import not at the top and some lambda assignments).

### Line Length

- `line-length = 88` within `tool.ruff`
- `max-line-length = 100` within `tool.ruff.lint.pydocstyle`
- `max-doc-length = 94` within `tool.ruff.lint.pydocstyle`

Note that `max-line-length = 100` is the maximum line length that the Ruff linter accepts, while `max-doc-length` is the maximum line length for docstrings. The `line-length = 88` option is intended for a formatter. In theory, if we add the Ruff formatter, it will attempt to break lines to stay within this boundary. In the future, `line-length` will serve as the default value, and `max-line-length` will act as a buffer if the formatter is unable to break lines accordingly.

### DIFF
--- a/CADETProcess/dataStructure/aggregator.py
+++ b/CADETProcess/dataStructure/aggregator.py
@@ -2,8 +2,6 @@ import numpy as np
 
 from .dataStructure import Aggregator
 
-import numpy as np
-
 
 class NumpyProxyArray(np.ndarray):
     """A numpy array that dynamically updates attributes of container elements."""
@@ -50,8 +48,8 @@ class NumpyProxyArray(np.ndarray):
         if not isinstance(obj, NumpyProxyArray):
             return np.asarray(obj)
 
-        self.aggregator = getattr(obj, 'aggregator', None)
-        self.instance = getattr(obj, 'instance', None)
+        self.aggregator = getattr(obj, "aggregator", None)
+        self.instance = getattr(obj, "instance", None)
 
     def __array_function__(self, func, types, *args, **kwargs):
         """
@@ -103,9 +101,9 @@ class SizedAggregator(Aggregator):
 
     def _expected_shape(self, instance):
         if self.transpose:
-            return self._parameter_shape(instance) + (self._n_instances(instance), )
+            return self._parameter_shape(instance) + (self._n_instances(instance),)
         else:
-            return (self._n_instances(instance), ) + self._parameter_shape(instance)
+            return (self._n_instances(instance),) + self._parameter_shape(instance)
 
     def _get_values_from_container(self, instance, transpose=False, check=False):
         value = super()._get_values_from_container(instance, check=False)
--- a/CADETProcess/dataStructure/aggregator.py
+++ b/CADETProcess/dataStructure/aggregator.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+from typing import Any, Optional
+
 import numpy as np
 
 from .dataStructure import Aggregator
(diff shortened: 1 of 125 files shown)


daklauss (commented):
[Pycodestyl (E,W)](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w) check for too much whitespaces or too many blank lines. [Pyflake F](https://docs.astral.sh/ruff/rules/#pyflakes-f) rules check for f-string and imports. [Pydocstyle D](https://docs.astral.sh/ruff/rules/#pydocstyle-d) checks for missing docstrings and docstring rules. D212 and D213 are rules that describes wheter the summary line is on the first or second dock string line, so we have to decide which one to use. D212 is the first line. D203 is a blank line before class docstring. D100 is a missing docstring in a public module which we actually usually have? So ignoring D100 is maybe not necessary.

daklauss:
I annotated as good as i could. Added many docstring some better than other. In a following issue we should think about formating and i gues with time there will be some annotations i missunderstood. Also some of my added docstrings were a bit lazy i guess. But we could always enforcer more rules there.

The ruff pipeline always checks our rules. For all tests, there will be some ignored rules. If the ruff check fails, it does not interrupt the other ones.

daklauss:
I decided for following linting rules:
E, F, W, D, ANN

E and W are typical PycodeStyle rules like using tab and or mixed with withespace, whitespaces between operator and so on.

F are pyflake rules, checking for imports, f strings, unused imports and so on.

D are a set of docstring rules. Docstrings will be enforced for ale public classes functions and methods and packages with the only exeption beeing modules. Excluded are D100, D212, D203. We enforce an descriptive imperative sentence at the start of every docstring. Module docstring are excluded because there are problems with sphinx used docstrings that do not fit the standard rules. So you need to exclude the docstring from ruff checking by using #noqa. 

ANN is enforced type checking for pretty much any public and private function and method. Rules ANN401 is excluded because there are many cases where Type Any is more convenient to use. Maybe we can exclude this one at a later time.

For all test Files the Docstring and Annotation Rules are disabled. as well as E731 and E402 (module import not at top and some lambda assignments).

CADETProcess/dataStructure/aggregator.py
```diff
@@ -1,14 +1,15 @@
 import numpy as np
+import numpy.typing as npt
 
+from typing import Any, Optional, Callable, Iterable, Mapping
```
schmoelder, line 4: when sorting imports, make sure to group by 
1. Standard Library
2. External libraries
3. Internal modules

daklauss: Good point.

CADETProcess/fractionation/fractionationOptimizer.py
```diff
@@ -80,38 +87,46 @@ def optimizer(self, optimizer):
...
+        """
+        Set up the Fractionator for optimizing the fractionation times of Chromatograms.
 
         Parameters
         ----------
         simulation_results: object
             Simulation results to be used for setting up the Fractionator object.
         purity_required : list of floats
             Minimum purity required for the components in the fractionation.
-        components: list, optional
```
schmoelder, line 103: this should be Optional[list]

PR #260, opened 2025-04-24
Use XDG Base Directory Specification for tmp and cache directories

ronald-jaepel:
Setting the location for the `optimizationProblem.cache_directory` and `settings.temp_dir` can be unwieldy, e.g. when running optimizations on alternating systems (such as the IBT servers or a personal device) or when running the example library on the IBT server (because the example library should not contain config information about the IBT server localstorage details).

This PR suggests using environment variables, specifically the XDG Base Directory specification to choose the tmp and cache folders.

I think this PR might cause problems if two optimization problems with the same name are run at the same time on the same system. Is that a use-case we want to allow?

### DIFF
--- a/CADETProcess/optimization/optimizationProblem.py
+++ b/CADETProcess/optimization/optimizationProblem.py
@@ -1,7 +1,9 @@
 import copy
 from functools import wraps
+import hashlib
 import inspect
 import math
+import os
 from pathlib import Path
 import random
 import shutil
@@ -15,6 +17,7 @@ import numpy as np
 import numpy.typing as npt
 
 from CADETProcess import CADETProcessError
+from CADETProcess.hashing import digest_string
 from CADETProcess import log
 from CADETProcess import settings
 
@@ -2759,7 +2762,13 @@ class OptimizationProblem(Structure):
     def cache_directory(self):
         """pathlib.Path: Path for results cache database."""
         if self._cache_directory is None:
-            _cache_directory = settings.working_directory / f'diskcache_{self.name}'
+            if 'XDG_CACHE_HOME' in os.environ:
+                _cache_directory = (
+                    Path(os.environ['XDG_CACHE_HOME']) / 'CADET-Process' /
+                    digest_string(settings.working_directory) / f'diskcache_{self.name}'
+                )
+            else:
+                _cache_directory = settings.working_directory / f'diskcache_{self.name}'
         else:
             _cache_directory = Path(self._cache_directory).absolute()
 
--- a/CADETProcess/settings.py
+++ b/CADETProcess/settings.py
@@ -13,7 +13,7 @@ This module provides functionality for general settings.
     Settings
 
 """
-
+import os
 from pathlib import Path
 import shutil
 import tempfile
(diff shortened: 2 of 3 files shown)


CADETProcess/optimization/optimizationProblem.py
```diff
@@ -2759,7 +2760,10 @@ def cached_steps(self):
     def cache_directory(self):
         """pathlib.Path: Path for results cache database."""
         if self._cache_directory is None:
-            _cache_directory = settings.working_directory / f'diskcache_{self.name}'
+            if 'XDG_CACHE_HOME' in os.environ:
+                _cache_directory = Path(os.environ['XDG_CACHE_HOME']) / 'CADET-Process' / 'cache' / f'diskcache_{self.name}'
```
schmoelder, line 2764: I guess we could leave out the `cache` subdir here since it's already specified as the `XDG_CACHE_HOME`.

CADETProcess/settings.py
```diff
@@ -120,7 +120,10 @@ def log_directory(self):
     def temp_dir(self):
         """pathlib.Path: Directory for temporary files."""
         if self._temp_dir is None:
-            _temp_dir = self.working_directory / 'tmp'
+            if 'XDG_CACHE_HOME' in os.environ:
+                _temp_dir = Path(os.environ['XDG_CACHE_HOME']) / 'CADET-Process' / 'tmp'
```
schmoelder, line 124: Here, we could use `XDG_RUNTIME_DIR`
>  There is a single base directory relative to which user-specific runtime files and other file objects should be placed. This directory is defined by the environment variable $XDG_RUNTIME_DIR.

schmoelder (commented):
Good idea, many thanks!

CADETProcess/settings.py
ronald-jaepel: `XDG_RUNTIME_DIR` seemed like it could bring complications with the requirements it holds:

> The directory MUST be owned by the user, and they MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

schmoelder:
> I think this PR might cause problems if two optimization problems with the same name are run at the same time on the same system. Is that a use-case we want to allow?

Good point. So either we need to check upfront or we create a unique name. 

I actually don't mind the unique name option (e.g. via a some UUID/time/hash suffix) since we don't really use the data after the optimization anyway. Also, the `tempdir` will be not be as relevant in the future since we will probably rely more and more on the `c-api`.

CADETProcess/optimization/optimizationProblem.py
```diff
@@ -2759,7 +2761,11 @@ def cached_steps(self):
     def cache_directory(self):
         """pathlib.Path: Path for results cache database."""
         if self._cache_directory is None:
-            _cache_directory = settings.working_directory / f'diskcache_{self.name}'
+            if 'XDG_CACHE_HOME' in os.environ:
+                start_time = int(datetime.now().timestamp() * 1000)
+                _cache_directory = Path(os.environ['XDG_CACHE_HOME']) / 'CADET-Process' / f'{self.name}_{start_time}'
+            else:
+                _cache_directory = settings.working_directory / f'diskcache_{self.name}'
```
schmoelder, line 2768: Should we keep the cache directory name consistent regardless whether `XDG_CACHE_HOME` is specified or not? I.e.  always use `f'{self.name}_{start_time}'`?

schmoelder: After our discussion online: let's use the following path:
```
Path(os.environ['XDG_CACHE_HOME']) / 'CADET-Process' / hash(settings.working_directory) / f'diskcache_{self.name}')
```
This has the same limitatations as the regular cache dir (e.g. don't run two optimizations with the same name in the same working directory) and does not pollute cache space by creating too many directories that never the overwritten.

PR #266, opened 2025-05-06
Split optional dependencies and dependency groups

daklauss:
Adding dependency groups and splitting optional dependencies so that optional dependencies like ax are actual additional optional features where as dependency groups are dependencies used for special use cases like for developing. To install dependency groups like dev in the future you have to use ```pip install CADET-process --group  dev```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,10 +40,8 @@ dependencies = [
 
 [dependency-groups]
 testing = [
-    "certifi", # tries to prevent certificate problems on windows
-    "pytest",
-    "pre-commit", # system tests run pre-commit
     "coverage",
+    "pytest",
     "pytest-cov"
 ]
 docs = [
@@ -56,14 +54,19 @@ docs = [
     "sphinx-sitemap>=2.5.0",
 ]
 dev = [
-    {include-group = "testing"},
     {include-group = "docs"},
+    {include-group = "testing"},
+    "certifi", # tries to prevent certificate problems on windows
+    "pre-commit",
     "ruff",
 ]
 
 [project.optional-dependencies]
 ax = [
     "ax-platform >=0.3.5,<1.0.0"
+]
+all = [
+   "CADET-Process[ax]",
 ]
 
 [project.urls]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -54,14 +54,22 @@ docs = [
...
     "sphinx-sitemap>=2.5.0",
 ]
-ax = [
-    "ax-platform >=0.3.5"
+dev = [
+    {include-group = "testing"},
+    {include-group = "docs"},
+    "ruff",
 ]
 coverage = [
```
schmoelder, line 62: Do we really need to split `testing` and `coverage`? How large is the difference in number of packages?

pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "scipy>=1.11",
 ]
 
-[project.optional-dependencies]
+[dependency-groups]
 testing = [
```
schmoelder, line 42: I don't think I would include `ax` here.

pyproject.toml
daklauss: I don't think this split is necessary. The only additional package in ```coverage``` are ```pytest-cov``` and ```coverage```.

pyproject.toml
daklauss: There is an option to run the coverage runs manually withing github actions with ```slow``` tests, so it would include every test. In this case, ```ax``` is necessary. If we wouldn't include ```ax``` here, we would need another dependency group and change the pipeline accordingly or otherwise ignore the option.

schmoelder: I see. We could also check whether ax is installed and only run the tests if it is installed. 
And in the CI we could simply install CADET-Process with all optional dependencies to ensure ax is installed.

daklauss: Yeah i also thought about that option.

daklauss: So i removed ax from the ```testing``` dependency group and added within the pipeline that ```ax``` as optional dependency should be installed. Do you mean by checking if all optional dependencies are installed for the user if he want's to run all tests without having ```ax``` installed locally?

schmoelder: It should skip the test if `ax` is not installed locally. Maybe raise a warning.

PR #268, opened 2025-05-13
Add Zonal Model Builder

AnanEzeiry:
## Todo
- [x] Rebase
- [x] Fix ruff / formatting
- [x] Add tests
- [x] Add documentation


CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+import math
+
+#Assuming one layer of membrane; uniform binding model
+binding_model = Langmuir(component_system)
+binding_model.is_kinetic = True
+binding_model.adsorption_rate = [9.5e-3]
+binding_model.desorption_rate = [1]
+binding_model.capacity = 1.2e3
+
+#In both scenarios axial or radial configuration, we defined one inlet that goes into one cstr
```
lieres, line 16: defined -> define

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+inlet = Inlet(component_system, name='inlet')
+volumetric_flow_rate = 1e-6
+inlet.flow_rate = volumetric_flow_rate #flow rate is constant assuming constant liquid density
```
lieres, line 19: flow rate could vary due to other reasons

CADETProcess/modelBuilder/ZRMBuilder.py
lieres: cstr -> CSTR

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+            if self.binding_model:
+                zone.binding_model = self.binding_model
+            zone.length = length
+            zone.total_porosity = porosity
+            zone.diameter = (4 * self.segments_area[i] / math.pi) ** 0.5 #from the segment area into diameter
+            zone.axial_dispersion = axial_dispersion
+            self.flow_sheet.add_unit(zone)
+            self.zones.append(zone)
+
+        # Create symmetric CSTRs' volumes
```
lieres, line 84: for later purpose: define CSTRs before and behind the membrane stack separately

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+            zone_fraction = self.segments_area[i] / remaining_area
+            next_cstr_fraction = 1 - zone_fraction
+            self.flow_sheet.set_output_state(
+                f'cstr{i+1}',
+                {
+                    f'cstr{i+2}': next_cstr_fraction,
+                    f'zone{i+1}': zone_fraction
+                }
+            )
+        # Inlet and outlet connections
```
lieres, line 126: This is the only place where axial flow and radial flow differs. You can already include both variants with conditional statements.

AnanEzeiry: It's not the previous block? I though this block is similar in both cases!

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+            self.flow_sheet.add_connection(self.cstrs[i], self.cstrs[i + 1])
+
+        # Back connections from zones to second half of CSTRs
+        for i in range(self.n - 1, -1, -1):
+            target_idx = 2 * self.n - i - 1
+            self.flow_sheet.add_connection(self.zones[i], self.cstrs[target_idx])
+
+        # Set output states for flow splitting
+        total_area = sum(self.segments_area)
+        epsilon = 1e-12  # Prevent divide-by-zero
```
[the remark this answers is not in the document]
AnanEzeiry: Epsilon, replaced by lowest_bound to avoid dividing by zero

CADETProcess/modelBuilder/AxialZRMBuilder.py
```diff
@@ -0,0 +1,171 @@
...
+from CADETProcess.processModel import FlowSheet, Process
+import math
+
+class ZRMBuilder:
+    def __init__(self, component_system, flow_rate, segments_area, configuration, init_c=0, binding_model=None, name=None):
+        """
+        Builder class for Axial Zoneal Rate Model (ZRM) with symmetric cstrs layout.
+
+        Args:
+            component_system: number of components.
```
lieres: The component system is more than just the number of components.

CADETProcess/modelBuilder/AxialZRMBuilder.py
```diff
@@ -0,0 +1,171 @@
...
+
+    @property
+    def cycle_time(self):
+        return self.process.cycle_time  
+    
+    
+    def Segments_ratio(self):
+        """Returns the surface area ratio of each zone."""
+        total_area = sum(self.segments_area)
+        if total_area == 0:
```
lieres: Line 71: Total area are cannot be 0 when all segment areas are >0.

CADETProcess/modelBuilder/AxialZRMBuilder.py
```diff
@@ -0,0 +1,171 @@
...
+            raise ZeroDivisionError("Total area of segments is zero.")
+        return [area / total_area for area in self.segments_area]
+
+    def build(self, length, porosity, axial_dispersion, volumes):
+        """Constructs the flow sheet."""
+        if len(volumes) not in (self.n, 2 * self.n):
```
lieres: Line 77: I would prefer two vectors for the CSTRs before and behind the membrane. If the second vector is empty, it is replaced by a copy of the first one. Otherwise, its length needs to be the same es of the first one.

CADETProcess/modelBuilder/AxialZRMBuilder.py
```diff
@@ -0,0 +1,171 @@
...
+        return self.flow_sheet
+
+    def _initialize_units(self, length, porosity, axial_dispersion, volumes):
+        """Initializes zones and cstr units in the flow sheet."""
+        self.flow_sheet.add_unit(self.inlet, feed_inlet=True)
+        self.flow_sheet.add_unit(self.outlet)
+
+        # Create membrane zones
+        for i in range(self.n):
+            zone = LumpedRateModelWithoutPores(self.component_system, f"zone{i+1}")
```
lieres: Line 94: If possible, allow to externally specify the transport model.

CADETProcess/modelBuilder/AxialZRMBuilder.py
AnanEzeiry: If I understand you correctly, I did that into the same class within cstr_in and cstr_out, when the problem is symmetric then the cstr_out will be specified  automatically

CADETProcess/modelBuilder/AxialZRMBuilder.py
AnanEzeiry: you mean that it's also include the type, e.g salt ?

lieres: Yes

CADETProcess/modelBuilder/AxialZRMBuilder.py
lieres: Yes, but I am referring to the interface.

AnanEzeiry: I see

CADETProcess/modelBuilder/AxialZRMBuilder.py
```diff
@@ -0,0 +1,171 @@
+from CADETProcess.processModel import ComponentSystem
+component_system = ComponentSystem(1) #number of components
+
+from CADETProcess.processModel import Langmuir
```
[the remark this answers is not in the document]
AnanEzeiry: What do you mean by “through the ZRMBuilder interface” (line 3)? What I thought is that the binding_model is specified as a ZRMBuilder input, and its parameters are specified externally, independent of ZRMBuilde.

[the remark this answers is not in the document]
lieres: That seems like a valid solution.

schmoelder:
Hi Anan, I added the description of the PR. Please proceed with the missing to-dos before we can merge this. Let us know if you need help.

CADETProcess/modelBuilder/AxialZRMBuilder.py
AnanEzeiry: Done

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Done

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Added

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+            name: Optional name for the builder instance.
+        """
+        self.component_system = component_system
+        self.inlet = inlet
+        self.outlet = outlet
+        self.flow_rate = flow_rate
+        self.segments_area = segments_area
+        self.n = len(segments_area)
+        self.binding_model = binding_model
+        self.name = name or 'AxialZRMBuilder'
```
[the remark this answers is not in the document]
AnanEzeiry: The current script addresses both axial and radial configurations

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,135 @@
...
+        self._initialize_units(length, porosity, axial_dispersion, volumes)
+        self._connect_units()
+        return self.flow_sheet
+
+    def _initialize_units(self, length, porosity, axial_dispersion, volumes):
+        """Initializes zones and CSTR units in the flow sheet."""
+        self.flow_sheet.add_unit(self.inlet, feed_inlet=True)
+        self.flow_sheet.add_unit(self.outlet)
+
+        # Create and configure zones
```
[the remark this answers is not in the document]
AnanEzeiry: Fixed

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Defined

CADETProcess/modelBuilder/AxialZRMBuilder.py
AnanEzeiry: Fixed

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+    def build(
+        self,
+        rate_model: type,
+        length: float,
+        porosity: float,
+        axial_dispersion: float,
+        volumes_in: list[float],
+        volumes_out: list[float]
+    ) -> FlowSheet:
+        """Construct the flow sheet."""
```
schmoelder: Improve docstrings

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        Convert text to a consistent case
+        to eliminate capitalization discrepancies.
+        checking whether the conf is axial or radial,
+        Otherwise rasing an error
+        """
+        if self.configuration not in ['axial', 'radial']:
+            raise ValueError("Configuration must be either 'axial' or 'radial'")
+
+    def build(
```
schmoelder:
```suggestion
    def build_flow_sheet(
```

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+    def cycle_time(self)  -> float:
+        """float: Get the cycle time of the process."""
+        return self.process.cycle_time
+
+    @cycle_time.setter
+    def cycle_time(self, time: float)  -> None:
+        self._cycle_time = time
+
+    @property
+    def binding_model(self) -> None:
```
schmoelder:
```suggestion
    def binding_model(self) -> BindingBaseClass:
```

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+            init_c: Union[int, float, list] = 0,
+            binding_model: Optional['BindingBaseClass'] = None,
+            name: Optional[str] = None
+        ) -> None:
+        """
+        Builder class for Axial/Radial Zonel Rate Model (ZRM).
+
+        with symmetric/Non-symmetric cstrs layout.
+
+        Args:
```
schmoelder: Be consistent with docstrings format. In CADET-Process, we use numpy style docstrings, not Google style.

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        self.cstrs_out = []
+
+        self.component_system = component_system
+        self.flow_rate = flow_rate
+
+        # Checking negative values for flow rate
+        self.segments_area = segments_area
+        self.init_c = init_c
+        self.configuration = configuration.lower()
+        self._validate_inputs()
```
schmoelder: Generally, it's fine to have a validate_inputs method. However, I would recommend checking before assigning attributes. That probably means, your `_validate_inputs` method should take arguments and return True/False or raise an exception.

This also makes it easier to test the individual method.

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+            segments_area: list[float],
+            configuration: str,
+            init_c: Union[int, float, list] = 0,
+            binding_model: Optional['BindingBaseClass'] = None,
+            name: Optional[str] = None
+        ) -> None:
+        """
+        Builder class for Axial/Radial Zonel Rate Model (ZRM).
+
+        with symmetric/Non-symmetric cstrs layout.
```
schmoelder: Fix formatting / Add more description. What are the different configurations / layouts.

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+class ZRMBuilder:
+    """
+    Class to build zones of compartmentalized  membrane chromotography device.
+
+    Attributes
+    ----------
+    segments_area : int
+        Surface areas of the segmented membrane.
```
schmoelder: Add missing attributes.

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        return self.process.cycle_time
+
+    @cycle_time.setter
+    def cycle_time(self, time: float)  -> None:
+        self._cycle_time = time
+
+    @property
+    def binding_model(self) -> None:
+        """
+        Get the binding model associated with this builder.
```
schmoelder: Check docstring conventions for properties.

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # Define the inlet/outlet, with filling their attributes
+        self.inlet = Inlet(component_system, name='inlet')
+        self.inlet.c = self.init_c
+        self.inlet.flow_rate = self.flow_rate
+
+        self.outlet = Outlet(component_system, name='outlet')
+
+    @property
+    def flow_sheet(self) -> FlowSheet:
```
schmoelder: do we need a flow_sheet property or should the ZRM only build a flow_sheet?

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        """FlowSheet: Return the FlowSheet instance associated with this builder."""
+        return self._flow_sheet
+
+    @property
+    def process(self)  -> Process:
```
schmoelder: do we need a process property or should the ZRM only build a process?

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        return self._process
+
+    @property
+    def cycle_time(self)  -> float:
+        """float: Get the cycle time of the process."""
+        return self.process.cycle_time
+
+    @cycle_time.setter
+    def cycle_time(self, time: float)  -> None:
+        self._cycle_time = time
```
schmoelder: Note, this setter does not automatically update the cycle time of the process.

schmoelder (commented):
We're slowly getting there. :nerd_face:

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+    def Segments_ratio(self) -> list[float]:
+        """Return the surface area ratio of each zone."""
+        total_area = sum(self.segments_area)
+        if total_area == 0:
+            raise ZeroDivisionError("Total area of segments is zero.")
+        return [area / total_area for area in self.segments_area]
+
+    def _validate_inputs(self) -> None:
+        """Validate init inputs."""
+        if isinstance(self.flow_rate, str):
```
schmoelder: Note, type / value checking of the flow rate is automatically performed by the `Inlet` when you set its property.

schmoelder: Just a note, if you were to test, don't only check for strings, but rather make sure it's a valid entry.

```suggestion
        if not np.isscalar(self.flow_rate):
```

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        ----------
+        binding_model : BindingBaseClass
+            The binding model to be set for the zones.
+        """
+        self._binding_model = binding_model
+        for membrane_segment in self.zones:
+            membrane_segment.binding_model = binding_model
+
+    # returning the ratio of each zone membrane area from the total membrane area
+    def Segments_ratio(self) -> list[float]:
```
schmoelder: Upper case variables are usually reserved for class definitions.

```suggestion
    def segments_ratio(self) -> list[float]:
```

CADETProcess/modelBuilder/ZRMFlowSheetBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError("All CSTR volumes (volumes_in) must be positive.")
+
+        if not volumes_out:
+            volumes_out = volumes_in
+        elif len(volumes_out) != self.n:
+            raise ValueError(f"Length of volumes_out must be {self.n}, got {len(volumes_out)}")
+        elif any(v_in < 0 for v_in in volumes_in):
+            raise ValueError("The cstr volume in volumes_in can not be negative float.")
+
+        # Proceed with the rest of the build logic here
```
schmoelder, line 150: Is this an open todo or can this be removed?

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+            volumes_out = volumes_in
+        elif len(volumes_out) != self.n:
+            raise ValueError(f"Length of volumes_out must be {self.n}, got {len(volumes_out)}")
+        elif any(v_in < 0 for v_in in volumes_in):
+            raise ValueError("The cstr volume in volumes_in can not be negative float.")
+
+        # Proceed with the rest of the build logic here
+
+        self._initialize_units(
+            rate_model,
```
schmoelder: Add type annotations

CADETProcess/modelBuilder/ZRMBuilder.py
```diff
@@ -0,0 +1,290 @@
...
+            length,
+            porosity,
+            axial_dispersion,
+            volumes_in,
+            volumes_out
+        )
+        self._connect_units()
```
schmoelder: Each method should do one thing only. Here the method implies that units are initialized, but instead you're connecting units (which is part of the flow sheet generation)

schmoelder: Update: this is a duplicate method. See below. Remove!

schmoelder (changes requested):

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Done

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Done

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Done

CADETProcess/modelBuilder/ZRMBuilder.py
AnanEzeiry: Yes, we need it since it is adjustable, for example, to add a PFR to account for the delay time

CADETProcess/modelBuilder/ZRMFlowSheetBuilder.py
AnanEzeiry: can you re-explain what do you mean here?

schmoelder:
We refactored the entire module. It's much simpler now and builds a (partial) flow sheet. It now also works for arbitrary column models, binding models, reaction models etc. This way, it's easier to integrate other unit operations, such as PFRs into flow sheet without having to hard code anything. Note, inlets and outlets now also need to be manually added after calling `build_flow_sheet`.

Since this is effectively now just a "meta-unit", in the future, we might want to consider an improved behavior in CADET-Process where we don't any "FlowSheetBuilder" but can simply resolve this when translating to CADET-Core. The same is true for other systems such as the `CarouselBuilder` or the `CompartmentBuilder`.

Note, tests have not yet been updated. :sweat_smile:

AnanEzeiry:
After refactoring the ZRMBuilder into the ZRMFlowSheetBuilder module, a new adjusted test was created, uploaded, and executed successfully. New documentation is required and is currently in progress. 😉

PR #279, opened 2025-06-06
Add tests for fractionation module

daklauss:
Added the Cases i used for the fractionating topic as test to have a small suit that checks wheter the standard ```fractionoptimizer``` and it's default parameter are suitable. Hopefully this shows possible instabilities with the cobyla optimizer when changing parameters.

I guess the cobyla fixture is a litle bit messy. 

Also changed the parameters for tol and rhoberg another time.

### DIFF
--- a/CADETProcess/fractionation/fractionationOptimizer.py
+++ b/CADETProcess/fractionation/fractionationOptimizer.py
@@ -65,8 +65,8 @@ class FractionationOptimizer:
         if optimizer is None:
             optimizer = COBYLA()
             optimizer.tol = 1e-4
-            optimizer.catol = 5e-3
-            optimizer.rhobeg = 1e-3
+            optimizer.catol = 1e-4
+            optimizer.rhobeg = 1e-4
 
         self.optimizer = optimizer
         self.log_level = log_level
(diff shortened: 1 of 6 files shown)


tests/test_fractionating.py
```diff
@@ -0,0 +1,296 @@
...
+    return (
+        1.0 / (np.sqrt(2.0 * np.pi) * sig) * np.exp(- exp) / 2
+    )
+
+
+class GaussianPulse(SolutionIO):
+    """Class to create a faked SolutionIO object with a Gauss function."""
+
+    def __init__(self, solution_struct, comp_system):
+        """solution struct has to be a dictionary with
```
schmoelder, line 24: Improve docstring

tests/test_fractionating.py
schmoelder, line 1: Consider a different name. "Fractionating" is not a really a thing. Consider what exactly you are testing here. Is it the FractionationOptimizer? Is it performance?

CADETProcess/fractionation/fractionationOptimizer.py
schmoelder, line 1: Does this make the optimization more robust or is this just "to play around"?

PR #283, opened 2025-06-12
Clean up __init__.py to increase import speed

schmoelder:
Currently, CADET-Process takes quite long to import. To increase import speed, I removed most import statements from the main `__init__.py`. Since we mostly import from the submodules anyways, this should not cause much trouble. AFAIK, only the `SimulationResults` were previously sometimes imported directly from `CADETProcess`.

Moreover, I udated most modules' `__all__` to reduce the number of imports when using wildcards (`*`)

```
python -X importtime -c "import CADETProcess"
```

Now give me ~500 ms, previously, this was ~2-3 s.

### DIFF
--- a/CADETProcess/__init__.py
+++ b/CADETProcess/__init__.py
@@ -16,30 +16,6 @@ __version__ = "0.11.0"
 # Imports
 from .CADETProcessError import *
 
-from . import log
-
 from .settings import Settings
 
 settings = Settings()
-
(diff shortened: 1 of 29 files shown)


hannahlanzrath (approved):
Awesome work! Looks good to me, tests are running on my Linux  and Windows as well and I can really notice the difference in some stuff I was running before.

Issue #105, opened 2024-03-14
Restructure `ReferenceBase`/`SolutionBase`


schmoelder:
- What interfaces *must* be implemented by which subclasses (e.g. resampling, smoothing, slicing etc.?)
- Alternatively, unify everything in `SolutionBase` and make use of internal dimensions.

schmoelder:
As mentioned, for this endeavor, we should also reconsider the resampling / interpolation methods of Solution objects. Currently, this is only possible for 1D solutions (i.e. SolutionIO). However, we might want to also do this with ND-objects. 

Here is a prototype that works for multiple dimensions:
```
import matplotlib.pyplot as plt
import numpy as np
from scipy.interpolate import RegularGridInterpolator
from scipy import stats

time = np.linspace(0, 10, 11)
axial_positions = np.linspace(0, 1, 5)
components = np.arange(0, 2)

dimensions = {
    'time': np.linspace(0, 10, 11),
    'axial_positions': np.linspace(0, 10, 11),
    # 'radial_positions': np.linspace(0, 10, 11),
    'channels': [0, 1, 2],
    'components': ['A', 'B'],
}

continuous_dimensions = [key for key, value in dimensions.items() if isinstance(value, np.ndarray)]
discrete_dimensions = [key for key, value in dimensions.items() if not isinstance(value, np.ndarray)]

fig, ax = plt.subplots()

c = np.zeros((len(time), len(axial_positions), len(components)))

for i in range(len(axial_positions)):
    solution = stats.norm.pdf(time, i + 3, 1)
    c[:, i, 1] = solution
    ax.plot(time, solution)

interpolators = []
for comp in components:
    interp = RegularGridInterpolator((time, axial_positions), c[..., comp], method='linear')
    interpolators.append(interp)

time_new = np.linspace(0, 10, 101)
axial_directions_new = np.linspace(0, 1, 11)

arrs = np.meshgrid(time_new, axial_directions_new, indexing='ij')

test_points = np.array([dim.ravel() for dim in arrs]).T

foo = interpolators[1](test_points).reshape(len(time_new), len(axial_directions_new))
for i in range(len(axial_directions_new)):
    solution = foo[:, i]
    ax.plot(time_new, solution, 'k--')
```

A challenge here is that we have independent (i.e. discrete) dimensions (e.g. components, channels) and continuous dimensions (space-time). While for interpolation, we only want to interpolate between the continuous dimensions, we still want to provide a uniform interface...

[...]

PR #289, opened 2025-06-25
Account for bound states in SolutionSolid

schmoelder:
This PR fixes an issue originally posted in our [Forum](https://forum.cadet-web.de/t/write-solution-solid-error-with-spreading-binding-model/1251).

Apparently, we had never tested the solid solution for binding models with multiple states.
Here, this is fixed by simply expanding the `component_coordinates`. 

Note, this is more of a quick-fix / hacky solution (no pun intended) which comes with some limitations. E.g. slicing the solution might not work properly.

Instead, we should consider something like a `BoundComponentSystem` which maps the bound states to the actual components. Maybe we could simply create a derived instance of the regular `ComponentSystem` or make use of the `Species`?

In any case, this is not very high priority until we have addressed some other issues:
- Multiple particle types (with individual binding models / bound states)
- General overhaul of solution class (@hannahlanzrath)

### DIFF
--- a/CADETProcess/solution.py
+++ b/CADETProcess/solution.py
@@ -1489,6 +1489,11 @@ class SolutionSolid(SolutionBase):
         return sum(self.bound_states)
 
     @property
+    def n_comp(self) -> int:
+        """int: Number of bound states."""
+        return self.n_bound
+
+    @property
     def ncol(self) -> int:
         """int: Number of axial discretization points."""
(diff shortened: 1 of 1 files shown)


hannahlanzrath (approved):
Good catch!

Looks fine to me, I added a reference in https://github.com/fau-advanced-separations/CADET-Process/issues/105.

PR #290, opened 2025-06-27
Add Badges

hannahlanzrath:
This PR adds badges for Release Version, CI status, PyPi Downloads per Month and the Zenodo DOI.

image


schmoelder:
Thanks for putting this together. I was just about to comment that adding links to the badges would be nice. But you were quicker than me! :nerd_face: 
I generally like it. Maybe, let's have a round of discussion on Monday about which badges we want to include in general.

PR #291, opened 2025-07-01
Use openID for PyPI instead of Token

hannahlanzrath:
This PR exchanges the usage of the API Token for using openID.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -3,32 +3,42 @@ name: Publish CADET-Process to PyPI
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
 jobs:
-  build-and-publish:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -1,34 +1,45 @@
 name: Publish CADET-Process to PyPI
 
 on:
-  release:
-    types: [published]
```
schmoelder, line 5: This pushes the release to PyPI only when we create a Github release.

.github/workflows/publish-to-pypi.yml
```diff
@@ -1,34 +1,45 @@
...
+  push:
+    tags:
+      - "v*"
```
schmoelder: This pushes the release to PyPI everytime we add a tag starting with "v".

PR #317, opened 2025-10-29
Add radial flow model

schmoelder:
AwwThis PR adds radial flow models to CADET-Process (see also: https://cadet.github.io/v5.1.0/interface/unit_operations/radial_flow_models.html#radial-flow-models-config).
Moreover, this adds a new framework to specify parameter-parameter dependencies (see also: https://cadet.github.io/master/interface/parameter_dependencies.html).
Note, until the generalized column model is available in CADET-Process, this only works for finite volume discretization.

## Todo
- [x] Also add LRM/LRMP
- [x] Add utility methods to calculate liquid / solid volume. Note, this is a bit more tricky because of the different geometry.
- [x] Add tests
- [x]  Anan -> Bug fix: Unnecessary ZeroDivisionError in unitOperation test
- [ ] Check coordinates in solution
- [ ] Check whether parameter dependencies also work for other unit operation


AnanEzeiry:
The following work has been done on a separate branch until we agree on the changes:

Radial flow has been implemented using the Radial prefix for the transport models and an inherited RadialFlowMixin class that provides radial geometry and hydrodynamic features. In addition, profile plotting features have been implemented for velocity and the Bodenstein number, and all tests are passing.

It is worth noting that this is a heuristic approach. A more substantial architectural change should be considered to prevent radial models from requiring unnecessary parameters.

PR #330, opened 2026-01-05
Improve plotting module

schmoelder:
This PR is a complete overhaul of the plotting module. It removes some methods that were not really relevant and unifies setting up figures, styling etc.

Note, it introduces several standard sizes: "1_col", "1.5_col", and "2_col". I got the settings from [here](https://www.elsevier.com/about/policies-and-standards/author/artwork-and-media-instructions/artwork-sizing) and I use "1_col" as a default. To be honest, it looks kinda small to me. Maybe this is because I use a High-DPI monitor? @hannahlanzrath do you have any ideas?

It still needs some testing but so far, it looks good.

### DIFF
--- a/CADETProcess/comparison/comparator.py
+++ b/CADETProcess/comparison/comparator.py
@@ -6,8 +6,6 @@ from typing import Any, Iterator, Optional
 import matplotlib.pyplot as plt
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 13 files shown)


hannahlanzrath:
Do you have an example file for me to quickly test the different col widths on my screen?

PR #331, opened 2026-01-05
Add k_eq to selected binding_models

schmoelder:
Note, this is currently just a property without setter.

We could also consider making this a default property for all binding models but we would first need to check whether all binding models really have a meaningful equilibrium constant.

Moreover, we could consider adding more properties such as Henry coefficients.

### DIFF
--- a/CADETProcess/processModel/binding.py
+++ b/CADETProcess/processModel/binding.py
@@ -197,6 +197,16 @@ class Linear(BindingBaseClass):
         "desorption_rate",
     ]
 
+    @property
+    def k_eq(self) -> list[float]:
+        """list[float]: Equilibrium constant."""
+        return np.divide(self.adsorption_rate, self.desorption_rate).tolist()
+
+    @property
+    def henry_coefficient(self) -> list[float]:
+        """list[float]: Henry coefficient."""
+        return self.k_eq
+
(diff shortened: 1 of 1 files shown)


CADETProcess/processModel/binding.py
```diff
@@ -197,6 +197,11 @@ class Linear(BindingBaseClass):
         "desorption_rate",
     ]
 
+    @property
+    def k_eq(self) -> list[float]:
+        """list[float]: Henry coefficients."""
```
hannahlanzrath: Why is k_eq here called "Henry coefficients" and in the Langmuir Equilibrium constant? Is this interchangeable in the Linear model?

schmoelder: Good question. In case of Langmuir, the Henry coefficient is the slope of the isotherm for c -> 0. I.e. for very small concentration, the Langmuir isotherm reduces to the linear model. 

To be fair, we could make this clearer in the wording and both add k_eq (k_ads / k_des) and Henry, which in case of linear yields the same result.

hannahlanzrath: To me that would be the more intuitive / consistent solution as an unsuspecting user hehe

PR #346, opened 2026-02-09
Fix: remove extra returns from docstring in get_callback

hannahlanzrath:
This PR removes the extra/doubled "Returns" from the docstring of the get_callback method, which locally caused errors in the documentation build.

I am unsure why this didn't error in the RTD build.


CADETProcess/optimization/scipyAdapter.py
hannahlanzrath, line 306: Returns is already set here.

PR #349, opened 2026-02-09
Check mass balance when evaluating stationarity

schmoelder:
This PR adds a check of the system mass balance when evaluating stationarity by comparing the integral sum of all system outlets with the feed mass.

Note, this might be an issue when reactions are involved since they do not conserve the mass of a component, but we would have to account for the conserved moiety (which is too much). Maybe we should consider making this optional, but since this is just the default Comparator used for the evaluation of stationarity, users can still write their own, so I don't think it's too bad for now.

In addition to the mass balance, I also had to rewrite some parts of the user solution times/section times. Previously, this was done to prevent Sundials struggling with very small time steps. However, this can also lead to other issues, e.g. with post-processing when section times do not align with the actual values of optimization variables. 
Note, this is still not stable and needs more testing.

### DIFF
--- a/CADETProcess/comparison/comparator.py
+++ b/CADETProcess/comparison/comparator.py
@@ -283,7 +283,7 @@ class Comparator(Structure):
         if ax is None:
             fig, axs = plotting.setup_figure(
                 **setup_figure_kwargs,
-                ncols=self.n_difference_metrics,
+                nrows=self.n_difference_metrics,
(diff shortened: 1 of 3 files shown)


schmoelder:
So far, I had no other issues with IDAS. I will merge and we can try to find another solution should this come up again. But rounding leads to too many other issues.

PR #368, opened 2026-03-27
Model builder improvements

schmoelder:
## Summary
- Do not copy columns in model builders as to not lose reference to the object.
- Consistent naming of event times across ModelBuilders

### DIFF
--- a/CADETProcess/modelBuilder/carouselBuilder.py
+++ b/CADETProcess/modelBuilder/carouselBuilder.py
@@ -1,5 +1,5 @@
+import copy
 import warnings
-from copy import deepcopy
 from functools import wraps
 from typing import Any, NoReturn, Optional, Union
 
@@ -332,8 +332,7 @@ class CarouselBuilder(Structure):
                 flow_sheet.add_unit(unit.inlet_unit)
                 flow_sheet.add_unit(unit.outlet_unit)
                 for i_col in range(unit.n_columns):
-                    col = deepcopy(self.column)
-                    col.component_system = self.component_system
+                    col = copy.copy(self.column)
                     col.name = f"column_{col_index}"
                     if unit.initial_state is not None:
                         col.initial_state = unit.initial_state[i_col]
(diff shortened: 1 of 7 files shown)


CADETProcess/modelBuilder/clrBuilder.py
```diff
@@ -34,7 +33,7 @@ def __init__(
         c_feed: list[float],
         flow_rate: float,
         feed_duration: float,
-        recycle_off: float,
+        t_recycle_off: float,
```
hannahlanzrath, line 36: Renaming this is a good idea

CADETProcess/modelBuilder/serialColumnsBuilder.py
```diff
@@ -150,14 +144,12 @@ def _build_flow_sheet(
         eluent_2 = Inlet(component_system, name="eluent_2")
         eluent_2.c = c_eluent
 
-        column_1 = copy.deepcopy(column)
+        column_1 = copy.copy(column)
```
hannahlanzrath: What purpose did the deepcopy serve before, why is it not necessary anymore?

hannahlanzrath (commented):
Looks good, I assume column doesnt need a deepcopy here, since it doesnt have a nested child object but only top level attributes?

CADETProcess/modelBuilder/carouselBuilder.py
```diff
@@ -332,8 +332,7 @@ def _add_units(self, flow_sheet: FlowSheet) -> None:
                 flow_sheet.add_unit(unit.inlet_unit)
                 flow_sheet.add_unit(unit.outlet_unit)
                 for i_col in range(unit.n_columns):
-                    col = deepcopy(self.column)
-                    col.component_system = self.component_system
+                    col = copy.copy(self.column)
```
schmoelder: check if we need to copy binding model.

CADETProcess/modelBuilder/serialColumnsBuilder.py
schmoelder: Deepcopy also creates new objects for all references which can lead to issues, especially with the component system.
However, we should check what to do with the binding model.

CADETProcess/modelBuilder/carouselBuilder.py
schmoelder: done, we now have an option for shallow copy (same binding model for all columns, default) or "own" copy.

PR #369, opened 2026-03-27
Add `ProcessMeta` dataclass to store process meta information

schmoelder:
`ProcessMeta` allows this information to be safely preserved when tweaking fractionation. For example, when a post-processing method determines cycle time from a chromatogram, the eluent and cycle time need to be adjusted to accurately compute KPIs.

### DIFF
--- a/CADETProcess/fractionation/fractionator.py
+++ b/CADETProcess/fractionation/fractionator.py
@@ -12,7 +12,7 @@ from CADETProcess.dataStructure import String
 from CADETProcess.dynamicEvents import Event, EventHandler
 from CADETProcess.fractionation.fractions import Fraction, FractionPool
 from CADETProcess.performance import Performance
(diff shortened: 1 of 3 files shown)


CADETProcess/fractionation/fractionator.py
```diff
@@ -210,18 +213,12 @@ def cycle_time(self) -> float:
...
-
-        Warning: This is not a robust feature! Side effects can ocurr in the Process!
-
         See Also
         --------
         productivity
 
         """
-        return self._cycle_time
+        return self.time[-1]
```
hannahlanzrath, line 221: Can you help me understand this change here?

hannahlanzrath (commented):
Everything looks good, I just dont understand the one change from `_cycle_time` to `time` but that's probably on me!

PR #370, opened 2026-03-27
fix: buffer capacity improvements

schmoelder:
## Summary
- Support different notations for protons in buffer capacity calculations
- Always set axis labels when plotting

### DIFF
--- a/CADETProcess/equilibria/buffer_capacity.py
+++ b/CADETProcess/equilibria/buffer_capacity.py
@@ -23,6 +23,7 @@ def preprocessing(
     buffer: npt.ArrayLike,
     pH: Optional[float | np.ndarray] = None,
     components: Optional[list] = None,
+    proton_index: int | None = None,
 ) -> tuple[dict[str, list], dict[str, np.ndarray], np.ndarray, dict[str, int], bool]:
     """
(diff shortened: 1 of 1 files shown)


CADETProcess/equilibria/buffer_capacity.py
```diff
@@ -68,10 +71,17 @@ def preprocessing(
             if comp not in components:
                 indices.pop(comp)
 
-    try:
-        proton_index = indices.pop("H+")
-    except ValueError:
-        raise CADETProcessError("Could not find proton in component system")
+    if proton_index is None:
+        proton_aliases = ("H^+", "H+", "$H^+$")
```
hannahlanzrath, line 75: Convinient addition!

hannahlanzrath (approved):

CADETProcess/equilibria/buffer_capacity.py
schmoelder: haha, thanks, but it might also be unpythonic.

> There should be one-- and preferably only one --obvious way to do it.

We should think about how to specify such things. It becomes more important when we finally add proper pH tooling or other tools where we want to specify *which* component has some specific role.

PR #273, opened 2025-05-22
Add parameter_space module

schmoelder:
See also #272 

## Todo
- [x] Bounds
- [x] Parameter normalization
- [x] Linear constraints
- [x] Linear equality constraints
- [~] Parameter dependencies
- [ ] Parameter mapping (i.e. how to map parameters to actual model parameters, providing different setter methods etc.)
- [ ] Mapping with indices, see also: [forum post](https://forum.cadet-web.de/t/specifying-indices-of-multidimensional-parameters-for-events-and-optimizationvariables/755) 
    - [ ] multidimensional arrays
    - [ ] `np.s_`
    - [ ] polynomial parameters
- [ ] Modularize parameter_space.py 
- [ ] Modularize tests for parameter_space.py
- [ ] Add NormalizedParameterSpace class
- [ ] Sampling methods (e.g. with hopsy)

## Open questions
- [x ] Where does normalization "live"? Parameter vs ParameterSpace? -> Parameter (e.g. choice parameters cannot be normalized anyway)
- [x] Where do dependencies "live"? Parameter vs ParameterSpace? Currently, they live in the Parameter. But we could consider moving them to the Parameter space to keep the Parameter class simple.
- [ ] Do we want to allow a more "dynamic" configuration? E.g. add_dependencies after instantiation?

PR #388, opened 2026-06-12
Parameter space + evaluation pipeline refactor

### LINKED ISSUE #272: Restructure optimization_problem module
The `OptimizationProblem` has become a [God Object](https://en.wikipedia.org/wiki/God_object). It does too many things at once, making it hard to maintain, test, extend, and reuse. For instance, for surrogate modeling, we need very similar functionality regarding the parameter space and evaluation pipelines.

To simplify this situation and make the code more modular, we want to split it into several modules:
- `parameter_space`: A module to define parameter spaces, including parameter names, bounds, and linear constraints, as well as normalization and parameter dependencies. Moreover, it should include functionality to "map" parameters to object attributes. Open question: Should we also create a module for the output spaces (e.g., objectives / nonlinear constraints)?
- `evaluation_pipeline`: A module to allow arbitrary computation sequences (e.g., configure process, run simulation, do some post-processing). For this purpose, we want to use [Pipefunc](https://pipefunc.readthedocs.io/en/latest/).

To not break existing functionality, we first want to create those new modules in isolation s.t. we can properly test them. Then, we want to incrementally start including them in the `OptimizationProblem`.

schmoelder:
Extracts `ParameterSpace` and `EvaluationPipeline` from `OptimizationProblem`, making the execution graph explicit and structurally shared rather than implicit via cache hits.

`OptimizationProblem` becomes a policy layer: it decides what is an objective, constraint, or callback, and delegates parameter semantics and evaluation to the two new modules. The old `optimizationProblem.py`, `cache.py`, and `diskcache.py` are removed; the `diskcache` dependency is dropped and replaced by pipefunc's built-in LRU/disk cache backends.

This enables surrogate modeling (a surrogate approximates `x → outputs` and slots in as a swappable backend) and DAG-level parallelism within a single individual's evaluation. Disk-backed caching across restarts is now available via `cache_directory`.

Supersedes #273 
Closes #272


schmoelder:
Thanks @talasunna for enduring my ramblings in the early prototypes of this huge refactor!

talasunna:
@schmoelder is this officially done now? Happy to have helped!

PR #391, opened 2026-06-17
Fix duplicate solution entries for multi-port units

### LINKED ISSUE #390: MCT `nchannel=N` produces N spurious "cycles"
When testing #273, I found a bug that predates that PR and became apparent with

```markdown
> commit `20584205656ec88ad3368f074fe37a8169b8278a`
> Author: Johannes Schmölder <j.schmoelder@fz-juelich.de>
> Date: Mon Feb 23 09:19:58 2026 +0100
> "Fix flow_rate TimeLine for solution_complete"
> (`CADETProcess/simulationResults.py`, 57 insertions / 37 deletions)
```
The original bug was introduced with the addition of ports

```markdown
> commit `0f8783db542d6eb876e083c93fd152fc04934b50`
> Author: daklauss <d.klauss@fz-juelich.de>
> Date: Fri Jul 12 21:19:07 2024 +0200
> "Add ports to cadetAdapter"
```

That commit added MCT/port support to `Cadet.get_simulation_results` in `cadetAdapter.py` and restructured the solution-reading code to loop `for port in unit.ports:` around the per-cycle read. The bulk/particle/solid/volume solution arrays are not actually port-specific (only inlet/outlet are), but the reads for them were nested *inside* that per-port loop. So for any unit with N ports (i.e. an MCT with `nchannel=N`) the same full bulk array gets read and appended once per port, producing N duplicate "cycle" entries in `solution_cycles` instead of one.

This duplication has always been wrong, but it was not apparent before commit 20584205656ec88ad3368f074fe37a8169b8278a. The old `_merge_cycles` logic in `simulationResults.py` bounded its merge loop by `self.n_cycles` (default 1):

```python
for i in range(1, self.n_cycles):   # range(1, 1) is empty
    solution_complete = np.vstack((solution_complete, cycles[i].solution[1:]))
```

### Bug Description

For an `MCT` unit operation with more than one channel (`nchannel=N`, N>1), accessing
`SimulationResults.solution` raises:

`ValueError: Expected size (T, ncol, N, n_comp)`

the first time it's accessed (e.g. via `simulation_results.solution[unit_name].bulk.solution`,
or anything that touches it, like `utils.interpolate_cb_from_simulation_results`).

### Steps to Reproduce

```markdown
import numpy as np
from CADETProcess.processModel import Inlet, MCT, Outlet, ComponentSystem, FlowSheet, Process
from CADETProcess.simulator import Cadet


def build_and_run(nchannel: int):
    cycle_time = 8400.0

    component_system = ComponentSystem(1)
    flow_sheet = FlowSheet(component_system)

    inlet = Inlet(component_system, name='inlet')
    mct1 = MCT(component_system, nchannel=nchannel, name='mct1')
    outlet = Outlet(component_system, name='outlet')

    mct1.discretization.ncol = 30

    flow_sheet.add_unit(inlet)
    flow_sheet.add_unit(mct1)
    flow_sheet.add_unit(outlet)

    flow_sheet.add_connection(inlet, mct1, destination_port='channel_0')
    flow_sheet.add_connection(mct1, outlet, origin_port='channel_0')

    process = Process(flow_sheet, 'MRE')

    mct1.length = 0.05
    mct1.channel_cross_section_areas = [1] * nchannel
    mct1.axial_dispersion = 1e-9
    mct1.solution_recorder.write_solution_bulk = 1
    mct1.exchange_matrix = np.zeros((nchannel, nchannel, 1))  # no actual coupling

    process.cycle_time = cycle_time

    inlet.flow_rate = [1.5e-4, 0, 0, 0]
    process.add_event('Base', 'flow_sheet.inlet.c', [1], 0)
    process.add_event('Start', 'flow_sheet.inlet.c', [1], 100)
    process.add_event('Stop', 'flow_sheet.inlet.c', [0], 460)

    simulator = Cadet()
    sim_results = simulator.simulate(process)

    cycles = sim_results.solution_cycles[mct1.name]['bulk']
    print(f"nchannel={nchannel}: num 'cycles' found = {len(cycles)} (expected: 1)")
    for i, c in enumerate(cycles):
        print(f"    cycle {i}: time[0]={c.time[0]}, time[-1]={c.time[-1]}, len={len(c.time)}")

    try:
        shape = sim_results.solution[mct1.name].bulk.solution.shape
        print(f"    -> solution access SUCCEEDED, shape={shape}")
    except ValueError as e:
        print(f"    -> solution access FAILED: {e}")


if __name__ == '__main__':
    for n in (1, 2, 3, 4):
        build_and_run(n)
        print()
```

### CADET-Process Version

commit 20584205656ec88ad3368f074fe37a8169b8278a

hannahlanzrath:
fixes #390 

For units with N ports (currently only MCT with nchannel=N), the bulk/particle/solid/volume solution reads were nested inside the per-port loop in Cadet.get_simulation_results, even though those arrays are not port-specific. This caused the same solution to be read and appended once per port, producing N duplicate entries in solution_cycles instead of one.

This was previously not breaking because the old cycle-merging logic in SimulationResults was bounded by self.n_cycles (default 1), so the duplicates were never touched. A later fix to that merging logic (20584205) correctly removed that coupling, which exposed this latent bug as a ValueError when accessing SimulationResults.solution for any MCT unit with nchannel > 1.

### DIFF
--- a/CADETProcess/simulator/cadetAdapter.py
+++ b/CADETProcess/simulator/cadetAdapter.py
@@ -487,281 +487,32 @@ class Cadet(SimulatorBase):
         try:
             solution = Dict()
             for unit in process.flow_sheet.units:
-                solution[unit.name] = defaultdict(list)
-
-                port_flag = unit.has_ports
-
-                if port_flag:
-                    solution[unit.name]["inlet"] = defaultdict(list)
-                    solution[unit.name]["outlet"] = defaultdict(list)
-
                 unit_index = self.get_unit_index(process, unit)
-                unit_solution = cadet.root.output.solution[unit_index]
-
-                unit_coordinates = cadet.root.output.coordinates[unit_index].copy()
-                particle_coordinates = unit_coordinates.pop('particle_coordinates_000', None)
-
-                for port in unit.ports:
-                    port_index = self.get_port_index(process.flow_sheet, unit, port)
-
-                    flow_in = process.flow_rate_timelines[unit.name].total_in[port]
-                    flow_out = process.flow_rate_timelines[unit.name].total_out[port]
-
-                    start = 0
-                    for cycle in range(self.n_cycles):
-                        end = start + len(time)
-
-                        if (
-                            f"solution_inlet_port_{port_index:03d}"
-                            in unit_solution.keys()
-                        ):
-                            sol_inlet = unit_solution[
-                                f"solution_inlet_port_{port_index:03d}"
-                            ][start:end,]
-                            if port_flag:
-                                solution[unit.name]["inlet"][port].append(
-                                    SolutionIO(
-                                        unit.name,
-                                        unit.component_system,
-                                        time,
-                                        sol_inlet,
-                                        flow_in,
-                                    )
-                                )
-                            else:
-                                solution[unit.name]["inlet"].append(
-                                    SolutionIO(
-                                        unit.name,
-                                        unit.component_system,
-                                        time,
-                                        sol_inlet,
-                                        flow_in,
-                                    )
-                                )
-
-                        if (
-                            f"solution_outlet_port_{port_index:03d}"
-                            in unit_solution.keys()
-                        ):
-                            sol_outlet = unit_solution[
-                                f"solution_outlet_port_{port_index:03d}"
-                            ][start:end, :]
-                            if port_flag:
-                                solution[unit.name]["outlet"][port].append(
-                                    SolutionIO(
-                                        unit.name,
-                                        unit.component_system,
-                                        time,
-                                        sol_outlet,
-                                        flow_out,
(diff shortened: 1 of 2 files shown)


schmoelder:
good catch. wanna add a test?

hannahlanzrath:
> good catch. wanna add a test?

Upon taking a closer look `test_compare_solution_shape` should have caught that after 20584205, but since it is a `slow` test, probably was not run. At least it fails on my machine after that commit and also for the latest release.

I have some suggestions:
- Similar to how I did it in the Metapackage, I can create a more comprehensive release workflow, that runs all test including the `slow` ones when a release is created on GitHub
- We could schedule an automatic workflow that runs once a month or so that runs all test including the `slow` ones

What do you think?

PR #393, opened 2026-06-19
CI: add schedule for slow tests

### LINKED ISSUE #392: Create nightly slow test run on dev
fixed by #393

hannahlanzrath:
fixes #392 

This uses cronjob to schedule a nightly run of all tests, including slow tests.

### DIFF
--- a/.github/workflows/scheduled-tests.yml
+++ b/.github/workflows/scheduled-tests.yml
@@ -1,0 +1,71 @@
+name: scheduled-tests
+
+on:
+  schedule:
+    - cron: "0 2 * * *"
+  workflow_dispatch:
+
+jobs:
+  test-job:
+    runs-on: ${{ matrix.os }}
+
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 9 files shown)


schmoelder:
thanks, @hannahlanzrath 
OOC, we know that we have some breaking "slow" tests, which are partially not "our fault" (e.g. due to cobyla breaking bounds). should we exclude those?

hannahlanzrath:
> thanks, @hannahlanzrath OOC, we know that we have some breaking "slow" tests, which are partially not "our fault" (e.g. due to cobyla breaking bounds). should we exclude those?

I think so, should I add another pytest tag for those?

schmoelder:
> I think so, should I add another pytest tag for those?

For external tool bugs (COBYLA bounds violations): use `pytest.mark.xfail(reason="COBYLA may violate bounds; upstream issue", strict=False)`. This documents the known failure, runs the test, passes CI, but alerts you via `XPASS` if the external tool ever fixes it. `strict=False` means an unexpected pass doesn't break CI either.

For internal bugs: let's create issues and decide together whether to defer the nightly launch or if we want to exclude them.

PR #395, opened 2026-06-24
Fix deepcopy failure with scipy 1.18 interpolation objects

schmoelder:
scipy 1.18.0 added `_xp` and `_xp_internal` (Array API dispatch) as module-object instance attributes on `PPoly`, `PchipInterpolator`, and related classes. Module objects are not picklable, so `copy.deepcopy` now fails for any object that transitively holds a scipy interpolator — breaking `ReferenceIO` construction, `SolutionIO.derivative`, and fractionation tests. `TimeLine._piecewise_poly` and `SolutionIO._solution_interpolated`/`_dm_dt_interpolated` are derived or lazy caches that are excluded from pickle state via `__getstate__`/`__setstate__`; both are recomputed on next use. scipy is capped at `<1.18` so dependabot opens a PR when 1.18 support can be restored. A MRE and upstream issue report for scipy are pending.

### DIFF
--- a/CADETProcess/solution.py
+++ b/CADETProcess/solution.py
@@ -287,6 +287,17 @@ class SolutionIO(SolutionBase):
         self.update_transform()
         self.flow_rate = self.flow_rate.slice(self.time[0], self.time[-1])
 
+    def __getstate__(self) -> dict:
+        """Return pickle state, excluding interpolation caches."""
+        state = self.__dict__.copy()
+        state["_solution_interpolated"] = None
+        state["_dm_dt_interpolated"] = None
+        return state
+
+    def __setstate__(self, state: dict) -> None:
+        """Restore from pickle state; caches are recomputed lazily on next access."""
+        self.__dict__.update(state)
+
     @property
     def derivative(self) -> "SolutionIO":
         """SolutionIO: Derivative of this solution."""
--- a/CADETProcess/dynamicEvents/section.py
+++ b/CADETProcess/dynamicEvents/section.py
@@ -326,6 +326,18 @@ class TimeLine:
         """list: scipy.interpolate.PPoly for each dimension."""
         return self._piecewise_poly
 
+    def __getstate__(self) -> dict:
+        """Return pickle state, excluding the derived piecewise polynomial cache."""
+        return {"_sections": self._sections}
+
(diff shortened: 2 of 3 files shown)


CADETProcess/comparison/difference.py
```diff
@@ -192,6 +192,10 @@ def reference(self, reference: SolutionBase) -> None:
                 f"Expected types: {self._valid_references}."
             )
 
+        # Defensive copy before transforms are applied. This is redundant because
```
hannahlanzrath: This is just an added comment, right?

CADETProcess/solution.py
```diff
@@ -287,6 +287,17 @@ def update_solution(self) -> None:
...
 
+    def __getstate__(self) -> dict:
+        """Return pickle state, excluding interpolation caches."""
+        state = self.__dict__.copy()
+        state["_solution_interpolated"] = None
+        state["_dm_dt_interpolated"] = None
+        return state
+
+    def __setstate__(self, state: dict) -> None:
+        """Restore from pickle state; caches are recomputed lazily on next access."""
```
hannahlanzrath, line 298: What does lazily mean here?

hannahlanzrath (commented):
I am not super familiar with pickling, it is when a python object gets translated back into machine code so it can be used for example in C++, is that right?

CADETProcess/comparison/difference.py
schmoelder: yes, it's really a follow-up issue where we *use* `deepcopy` which causes the issue. In analyzing the root cause, I found that maybe we could optimize our code a bit. Admittedly, a comment is not ideal, I will remove it.

CADETProcess/solution.py
schmoelder: some of our "private" attributes are really just caches (e.g. `_solution_interpolated`). So when pickling, we shouldn't worry too much about those, we invalidate the cache so they are recomputed when needed.

schmoelder:
"pickling" is about serializing objects, e.g. to disk but also between processes. I.e. we take all of the information that makes up the object and store it somewhere so we can recreate it.

hannahlanzrath (approved):

PR #398, opened 2026-06-29
Add BoFire Bayesian optimization adapter

schmoelder:
Adds `BoFire`, a Bayesian optimizer built on BoFire's ask/tell interface. `SoboStrategy` handles single-objective problems; `MoboStrategy` and `QparegoStrategy` handle multi-objective problems. Nonlinear constraints are modeled as additional GP outputs with sigmoid objectives rather than hard rejections, so the surrogate learns feasibility structure rather than discarding infeasible candidates. Linear constraints are passed directly to the BoFire `Domain`. The GP kernel is selectable; `"auto"` uses BoFire's default. Install via `pip install CADET-Process[bofire]`.

### DIFF
--- a/CADETProcess/optimization/__init__.py
+++ b/CADETProcess/optimization/__init__.py
@@ -67,6 +67,14 @@ Ax
    GPEI
    NEHVI
 
+BoFire
+------
+
+.. autosummary::
+   :toctree: generated/
+
+   BoFire
+
 Population
 ==========
 .. autosummary::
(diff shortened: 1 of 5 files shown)


kaanatak:
I tested the new features on a small optimization example and everything worked as expected. Also, the installation worked without any issues. I plan to use the Bayesian optimization functionality in my own project in the future. I am sure I will have more feedback and suggestions once I use it more extensively in larger problems. Currently, everything looks fine on my side.

PR #402, opened 2026-07-01
Update Yamamoto Documentation and add Yamamoto tests

hannahlanzrath:
Good news! Yamamoto is correctly implemented in CP. I added some tests to summarize a little what I learned during the past day and to have a benchmark for what accuracies to expect.

## Summary

- **`tests/tools/test_yamamoto.py`**: adds tests for the Yamamoto parameter
  estimation method using three benchmark cases:
  - *Rüdt et al. 2015, 200 mm LRMP, single protein*: shows that ν and k_eq are both
    recovered within 5 % when the column has enough theoretical plates (N≈2440).
  - *Püttmann et al. 2013, 14 mm GRM, three proteins*: documents the realistic accuracy
    limit of a short column and limiting pore-diffusion: ν within ~1 %, k_eq within ~25 %, limited by
    plate count (N≈70).
  - *Modified Püttmann et al. 2013, 200 mm LRMP, three proteins*: shows that k_eq accuracy
    improves to ~8 % on a longer column (N≈1000).

- **`docs/source/references.bib`** : adds `Yamamoto1983` and `Rudt2015`
  bibliography entries.

- **`examples/characterize_chromatographic_system/Yamamoto_method.md`**:  Extends the documentation with literature citations and a theory section.

### DIFF
--- a/examples/characterize_chromatographic_system/Yamamoto_method.md
+++ b/examples/characterize_chromatographic_system/Yamamoto_method.md
@@ -24,8 +24,40 @@ sys.path.append(root_dir.as_posix())
 
 # The Yamamoto method
 
-This example demonstrates how to estimate SMA binding parameters based on multiple gradient elution chromatograms
-using the Yamamoto method.
+The Yamamoto method {cite}`Yamamoto1983` estimates the two key SMA binding parameters, **characteristic charge $\nu$** and **equilibrium constant K**, from a small set of linear salt-gradient elution (LGE) experiments, without requiring an inverse problem to be solved.
+
+## Theory
+
+For a protein eluting under a linear salt gradient, the **normalized gradient slope** $GH$ and the **peak salt concentration** $I_R$ are related by {cite}`Rudt2015`:
+
+$$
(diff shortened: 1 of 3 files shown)


examples/characterize_chromatographic_system/Yamamoto_method.md
```diff
@@ -24,8 +24,40 @@ sys.path.append(root_dir.as_posix())
 
 # The Yamamoto method
 
-This example demonstrates how to estimate SMA binding parameters based on multiple gradient elution chromatograms
-using the Yamamoto method.
+The Yamamoto method {cite}`Yamamoto1983` estimates two key SMA binding parameters: the **characteristic charge ν** and the **equilibrium constant K**, from a small set of linear salt-gradient elution (LGE) experiments, without requiring full chromatogram fitting.
```
schmoelder: Use dollar math instead of Unicode for greek (and other math) symbols

schmoelder: also, maybe instead of a "without requiring full chromatogram fitting." write something like solving an inverse problem?

schmoelder (changes requested):
just some minor changes.

hannahlanzrath:
> just some minor changes.

Thank you, done!

PR #403, opened 2026-07-01
Wire callbacks through evaluation pipeline for cache reuse

### LINKED ISSUE #133: Handling of callbacks_dir
This is an issue to collect thoughts:

Calling:

`optimization_problem.evaluate_callbacks(ind)`

from outside of the optimization loop crashes if the callbacks use the `callbacks_dir` and are added without the `callbacks_dir` option in 
`optimization_problem.add_callback(callback, requires=[simulator], callbacks_dir="tmp")`.

Being able to hand `callbacks_dir` to `optimization_problem.evaluate_callbacks(ind)` would be useful.

schmoelder:
Callbacks re-ran the evaluator chain manually, duplicating simulation work already cached from objective evaluation for the same individual. `evaluate_callbacks` now calls `pipeline.evaluate()` for the chain, returning cached results when the same `x` was already evaluated, with a fallback to direct execution if the pipeline call fails.

`callbacks_dir` is now a proper parameter of `evaluate_callbacks` instead of a private attribute mutated onto `_MetricRecord` by the optimizer before each call. Per-callback directory resolution and `cleanup` both move into `evaluate_callbacks`, eliminating the side-channel and resolving the limitation from #133 where `callbacks_dir` could not be passed from outside the optimization loop.

`EvaluationPipeline.add_evaluator` gains a `cache` parameter (default `True`).

Also fixes several edge cases: `current_iteration="final"` no longer passed to `cleanup`; `callbacks_dir` coerced to `Path`; pipeline fallback logged at debug with traceback instead of silently swallowed; multi-eval-obj result indexing uses `id()`-based map; `frequency < 1` rejected at registration; `inspect.signature` guarded for non-standard callables; callback exception warnings preserve traceback via `exc_info=True`.

Closes #133.

### DIFF
--- a/CADETProcess/evaluation_pipeline/pipeline.py
+++ b/CADETProcess/evaluation_pipeline/pipeline.py
@@ -122,6 +122,7 @@ def _make_node(
     func: Callable,
     output_name: str,
     requires: list[str] | None,
+    cache: bool = True,
 ) -> PipeFunc:
     """Build a `PipeFunc` node with failure propagation and optional arg injection."""
     safe = _wrap_with_failure_propagation(func, stage=output_name)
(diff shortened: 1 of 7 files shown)


schmoelder:
Closing, as I directly commited the fixup on `dev`

PR #404, opened 2026-07-06
Pass parameter values by name in ParameterSpace and EvaluationPipeline

schmoelder:
In the optimization layer, a point in the search space was passed as a positional float vector: `set_values([0.5, 3.0])` relied on registration order to know which value belongs to which parameter. A float vector cannot hold a categorical value (`ChoiceParameter` could be registered but never written or sampled), and its limits leaked into everything built on it: the evaluation cache keyed on the raw float tuple (two vectors normalizing to the same integer created duplicate entries, and categorical choices could never be keyed at all), and integer variables needed ad-hoc rounding at each call site.

This PR makes the named assignment `{"length": 0.5, "resin": "A"}` the canonical input. `ParameterSpace.set_values` and the new `resolve` take a mapping and validate it strictly: unknown names, missing independent parameters, and caller-supplied dependent values raise. The positional vector remains for optimizers and polytope sampling, but as an explicit encoding: `TransformedSpace.encode`/`decode` convert between the two, with `decode` owning integer rounding and the categorical side-channel. `TransformedSpace` no longer has a write method, so every write goes through the one validated path. The pipeline cache is keyed by the `(name, value)` tuple, and `space.sample()` returns assignments, drawing categoricals uniformly from their valid values.

The public `OptimizationProblem` API is unchanged; optimizer vectors are decoded at the boundary. This completes the framework side of categorical support; the first optimizer that consumes categorical variables natively arrives with the BoFire adapter.

### DIFF
--- a/CADETProcess/evaluation_pipeline/pipeline.py
+++ b/CADETProcess/evaluation_pipeline/pipeline.py
@@ -152,8 +152,9 @@ class EvaluationPipeline:
     Parameters
     ----------
     space : ParameterSpace
-        Owns the evaluation objects and knows how to write a parameter vector
-        into them.  `evaluate` delegates to `space.set_values(x)`.
+        Owns the evaluation objects and knows how to write parameter values
+        into them.  `evaluate` decodes the vector and delegates to
+        `space.set_values`.
 
     Examples
     --------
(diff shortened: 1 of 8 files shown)


kaanatak (approved):
I tested this on my own cases and I was able to reproduce my results without any issues.

PR #405, opened 2026-07-07
Add sampling backends for ParameterSpace

schmoelder:
The hopsy-based construction was previously duplicated across `ParameterSpace.sample()`, `create_initial_values`, and `get_chebyshev_center`, making the implementation harder to maintain and preventing alternative sampling strategies from being plugged in. This PR consolidates that logic behind a sampler abstraction. `SamplerBase` now owns the complete feasibility pipeline—including bounds (with dependent parameters), linear constraints (including those involving dependent variables), integer rounding, and categorical sampling—while concrete backends are responsible only for generating candidate points.

`HopsySampler` remains the default for constrained parameter spaces, while the new `LatinHypercubeSampler` and `SobolSampler` provide space-filling designs for box-bounded surrogate modeling and DoE workflows. These designs are consumed in generation order, with categorical variables treated as additional stratified dimensions.

`get_chebyshev_center` now follows the same feasibility pipeline and returns either a verified feasible point or fails explicitly, rather than silently returning an infeasible one. `create_initial_values` gains a `sampler` argument, and a new user guide page documents `ParameterSpace` and its sampling interface independently.

The abstraction is intentionally extensible: additional backends, such as full-factorial grids, can be added by implementing `_candidates`.

### DIFF
--- a/CADETProcess/optimization/optimization_problem.py
+++ b/CADETProcess/optimization/optimization_problem.py
@@ -994,7 +994,6 @@ class OptimizationProblem:
             n_samples,
             seed=seed,
             include_dependent=False,
-            validate=lambda x: self.check_individual(x, check_nonlinear_constraints=False),
         )
         ts = self._space.transformed_space
         rows = []
--- a/CADETProcess/optimization/optimization_problem.py
+++ b/CADETProcess/optimization/optimization_problem.py
@@ -10,17 +10,14 @@ from __future__ import annotations
 import inspect
 import logging
 import math
-import random
 import shutil
 import warnings
 from collections.abc import Callable
 from pathlib import Path
 from typing import Any, Literal, Optional
 
-import hopsy
 import numpy as np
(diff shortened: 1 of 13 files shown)


kaanatak:
Thank you for the update! The sampling methods work as expected in my test cases. 

I have two suggestions for the Chebyshev center calculation:

1. Would it make sense for the Chebyshev center to be returned as a dictionary as well instead of an array? That would preserve the association between coordinates and their corresponding parameter names, making the result easier to interpret and less error-prone.
2. I noticed the center can look not intuitive for a "thin" parameter space. For example, with one parameter in bounds `(1, 600)` and another in `(1e-8, 1e-4)`, the center is approximately `(1, 5e-5)`. The second coordinate might make sense, but the first being near the lower bound seems a bit confusing. It may be the expected behavior since the largest sphere can shift along the thin dimension. If so, it might be worth mentioning this in the documentation.

PR #406, opened 2026-07-08
Add MetricSpace and proper Problem class

schmoelder:
This PR gives the output side of an optimization problem the same infrastructure the input side gained with `ParameterSpace`: `Metric` declares a named output, `MetricSpace` owns direction, constraint, and normalization nnotations, and `Problem` combines both spaces with a swappable `EvaluationBackend` into an optimizer-independent problem description that samplers and surrogates consume directly.
Building on that, `Population` stores parameters and metrics as named columns and derives everything else (directions, violations, feasibility, dominance) from the spaces instead of duplicating it per row.


kaanatak (commented):
Everything worked on my end. A few thoughts:

* In the diagnostics, it might be difficult to interpret the full constraint violation matrix. Maybe we could emphasize or explain it a bit more in the documentation.

* Should there be a uniqueness check when concatenating populations so that the same parameter sets are not included multiple times (or perhaps described as `run1`, `run2`, ...)? What do you think?

* Also, what would you think about calling problem evaluation from the `Population` as well? So, given a sample, run all evaluations (either serially or in parallel) and store the results by using a similar method to `evaluate()`. Or do you want to keep it separate from the pipeline and use `Population` solely as an organizer for results and post-processing?

schmoelder:
Hey Kaan, Thanks for the review!

On #1, agreed; added a section to `optimizer.md` explaining the `cv` matrix's block ordering, column counts, and sign convention, and pointing at `is_feasible`/`feasible`/`infeasible` instead of reading `cv` positionally.

On #2, I'd rather not dedup automatically in `concat`. Auto-deduplication would silently drop rows for callers who actually want repeated points (e.g. comparing evaluation noise across runs).

On #3, I'd rather not add evaluation as an instance method on `Population`. It'd mean `Population` holding a reference to a `Problem`/backend, which cuts against keeping it a pure results/organizer type.

kaanatak:
Thank you for the explanation! Sounds good to me!

schmoelder:
Hey @kaanatak I updated some more of the docs. I you have time, I'd appreciate some more feedback. :)
Also, have you tried building the docs via sphinx? If not, please try and let me know if you need help.

kaanatak:
Thanks for the update! I like it and found it easy to follow. Regarding building with Sphinx, I didn't encounter any issues. It takes slightly longer, but it works on my end.

I would have a minor suggestion. There exist a link from `Optimizer` to `Problem/Building a Population`. It might be nice to have a link from `Problem/Building a Population` to `Optimizer` to guide the users once more to the methods available within `Population` and for completeness.

schmoelder:
Thanks Kaan and good point, I added another sentence:

> The {ref}`optimizer_guide` documents the further accessors a `Population` exposes on optimization results, including the flat `f`/`g` projections, the constraint-violation matrices, and the feasibility filters.

Note, in the next step, we might restructure the entire subpackage but for now, I wanna merge it.

kaanatak:
Thanks for the update! Sounds good!

PR #413, opened 2026-07-21
Route evaluation-object fan-out through the DAG as an object axis

schmoelder:
Moves the per-evaluation-object loop out of `EvaluationPipeline.evaluate` and into the DAG as one ordered `object` axis, so a downstream node can consume all per-object results and reduce them. That is what makes aggregation expressible: a metric registered with `per_object=False` collects the whole array and returns one value, letting a multi-objective problem be reduced to a single scalar solvable by a cheaper single-objective optimizer (replacing the meta-score workaround). `per_object`/`mapspec` are available on objectives, constraints, meta scores, and callbacks; per-call `evaluation_objects=` subsets and `bypass_cache` are preserved; and there is now a single execution engine (the legacy per-object loop is deleted). Requires `pipefunc>=0.93.1`.


schmoelder:
Note, let's merge #434 first and then rebase this.

